PR #190, opened 2015-02-07
[gfs] rebinning_tau(), a function to replace change_mesh() from cthyb

krivenko:
Note: I do not use neither closest_point_pt, nor windowing(), because they do almost what I need, but not quite. Instead I wrote my own implementation of a simple linear interpolation algorithm.


parcollet:
Merged.

Issue #141, opened 2014-10-26
[1.2-dev] wrapper_desc_generator.py error


krivenko:
Hello,
I've been trying to regenerate `cthyb_desc.py` for cthyb. `wrapper_desc_generator.py` gives me the following error:

```
$ wrapper_desc_generator.py ../c++/solver_core.hpp -p -mpytriqs.applications.impurity_solvers.cthyb -o cthyb --moduledoc "The cthyb matrix solver"
Welcome to the wrapper desc file generator of TRIQS, based on libclang !
Parsing the C++ file (may take a few seconds) ...
... done. 
Extracting ...
Traceback (most recent call last):
  File "/home/igor/Physics/TRIQS/installed/bin/wrapper_desc_generator.py", line 60, in <module>
    functions, classes = parse(args.filename, debug = False, compiler_options = compiler_options, where_is_libclang = args.libclang_location)
  File "/home/igor/Physics/TRIQS/installed/bin/clang_parser.py", line 211, in parse
    raise RuntimeError, s + "\n... Your code must compile before making the wrapper !" 
RuntimeError: Clang reports the following errors in parsing
file /usr/include/limits.h line 123 col 16
'limits.h' file not found
... Your code must compile before making the wrapper !
```

Of course, my code is compiled well by `make`.
Here is what I see around line 123 in `/usr/include/limits.h`:

```
116  /* Get the compiler's limits.h, which defines almost all the ISO constants.
117 
118     We put this #include_next outside the double inclusion check because
119     it should be possible to include this file more than once and still get
120     the definitions from gcc's header.  */
121 #if defined __GNUC__ && !defined _GCC_LIMITS_H_
122 /* `_GCC_LIMITS_H_' is what GCC's file defines.  */
123 # include_next <limits.h>
124 #endif
```

My compiler:

```
$ clang++ -E -x c++ - -v < /dev/null
clang version 3.5.0 (tags/RELEASE_350/final)
Target: x86_64-pc-linux-gnu
Thread model: posix
Selected GCC installation: /usr/lib/gcc/x86_64-pc-linux-gnu/4.8.3
[...]

PR #191, opened 2015-02-07
[cmake] CMake module to detect LibClang location and compilation flags

krivenko:
See #141 for more details.


parcollet:
Merged.

PR #192, opened 2015-02-07
[arrays] tridiag_worker: support for complex Hermitian matrices

krivenko:
A tridiagonal complex Hermitian matrix can be easily (in linear time) made real symmetric through a similarity transformation. Such a transformation replaces the off-diagonal elements of the matrix by their absolute values. The transformation itself is given by a diagonal unitary matrix, which can be stored as a complex vector.

After an application of `dstev` to the real matrix, the eigenvectors must be back-transformed using the stored transformation matrix.


parcollet:
Merged.

PR #195, opened 2015-02-10
Cleaned up numeric_ops.hpp using types from std14

krivenko:
* Following today's discussion


krivenko:
> 1- why the leading typename ? I think it is useless, no ?

Ok, just forgot about these...

>  2- why the decay and the && ? You do not forward here, there is no need for forwarding reference,
why don't you simply use T const & ?

After running a couple of tests I convinced myself that `T const &` would, indeed, be enough. And those tests are included to the commit now.

parcollet:
If there is no forward, T&& is a useless complication...
Ok, I will squash and merge it later today.

parcollet:
Merged

PR #224, opened 2015-05-30
[cmake] New approach to running Python tests in TRIQS applications

krivenko:
This pull request is meant to improve the way Python tests of TRIQS applications are run.

# Problem description

To be able to successfully run a Python test script, the Python interpreter needs access to a 'logical sum' of two pytriqs package trees,

* one from an installation of the TRIQS library (`${TRIQS_PYTHON_LIB_DEST_ROOT}`);
* and one from the build directory of the application (normally `${CMAKE_BINARY_DIR}/python`).

The second tree takes precedence over the first in case both trees contain similar branches. 

# Current solution

There is currently a special CMake macro called `triqs_prepare_local_pytriqs_merged_with_my_python()`.
Besides other things, it scans through the installed pytriqs package tree and recreates the corresponding directory structure in `${CMAKE_BINARY_DIR}/pytriqs`.
Files in the installed tree are represented by symlinks in the destination tree. If a file has an counterpart in `${CMAKE_BINARY_DIR}/python`, the a symlink pointing at that location is created.
Once the merged tree is prepared, one can use it by adding `${CMAKE_BINARY_DIR}/pytriqs` to `PYTHONPATH`.

This solution proved to work, however, I've been seeing weird symlink-related build failures happening once in a while (especially, when I was developing two or more TRIQS applications simultaneously.)
So I decided to come up with a less complicated and fragile solution.

# Proposed solution

Here I define a new macro `triqs_prepare_local_pytriqs(python_destination)`. It has two functions.

* It creates three local wrapper scripts `build_pytriqs`, `build_ipytriqs` and `build_ipytriqs_notebook` (just as the old macro does). The scripts have received an update that puts them in sync with the scripts from `/shells` dir.
* It also generates a short python script `${CMAKE_BINARY_DIR}/sitecustomize.py`.

Python automatically imports `sitecustomize.py` before running a script or an interactive shell (given the CWD is `${CMAKE_BINARY_DIR}`). This scripts installs a special hook, which intercepts calls to the `__import__()` builtin function. The hook checks whether a part of the application subpackage (for instance, `pytriqs/applications/impurity_solvers/cthyb`) is about to be imported. If it is the case, the hook rewrites the name of the subpackage and effectively redirects import request to `${CMAKE_BINARY_DIR}/python`.

In the described approach no local copy of the pytriqs tree is needed at all.

For testing purposes I created a cthyb branch `new_python_build_system`, where the introduced mechanism is used: TRIQS/cthyb@c4ecb22cbece75539bf82a020fba798f11668690.

P.S. The old `triqs_prepare_local_pytriqs_merged_with_my_python()` is kept intact for the existing applications. It should be announced to be deprecated, if this pull request is accepted.

### DIFF
--- a/cmake/TRIQSConfig.cmake.in
+++ b/cmake/TRIQSConfig.cmake.in
@@ -34,6 +34,7 @@ set(TRIQS_LIBRARY_GSL     @GSL_LIBRARIES@)
 set(TRIQS_WITH_PYTHON_SUPPORT @TRIQS_WITH_PYTHON_SUPPORT@)
 set(TRIQS_HDF5_DIFF_EXECUTABLE @TRIQS_HDF5_DIFF_EXECUTABLE@)
 set(TRIQS_HDF5_COMMAND_PATH @TRIQS_HDF5_COMMAND_PATH@)
+set(TRIQS_PYTHON_LIB_DEST_ROOT @TRIQS_PYTHON_LIB_DEST_ROOT@)
 set(TRIQS_PYTHON_LIB_DEST @TRIQS_PYTHON_LIB_DEST@)
 
 # for people who want to quickly add everything TRIQS has detected...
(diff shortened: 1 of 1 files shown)


krivenko:
I'll need to extend and test this CMake macro more. Apparently, there is still a bug or two...

krivenko:
I've finished testing cthyb, dft_tools and Hubbard-I with this new patch.
Compilation, tests and installation - all run smoothly on my Linux machine.

krivenko:
@parcollet 
Dear Olivier,
we have discussed this new feature with Michel, and he does not mind merging the pull request.
Could anyone test it on Mac OS X? (Though I don't expect problems, since no compiled parts are involved).

parcollet:
I have discussed with Michel. The solution looks great to me.
I will test it on my mac, and then merge if no pb is found there.

krivenko:
I'm closing this PR, because its more up-to-date version now lives in the main repository as [new_python_test_system branch](https://github.com/TRIQS/triqs/commits/new_python_test_system).

PR #261, opened 2015-09-25
Mytriqs

lmessio:
In the doc : separation of the tutorials into python and c++
in det_manip : replacement of change_on_line_and_one_col by change_on_row_and_one_col, which works better


krivenko:
Squashed and merged as 488c9d8c33d94ca48ff965cae9821a13b74ffd28.

PR #282, opened 2016-01-05
Fixed bug in det_manip

JaksaVucicevic:
There is an underflow problem with det_manip because the determinant ratio returned from try_remove, try_remove2, etc. is in the end not calculated directly from inverse matrix elements, but from the actual ratio between the new and the old determinant. When the old determinant is underflowing (is zero), nan is returned.

This is now fixed by skipping the unnecessary step of recalculating the determinant ratio which we already have

newdet = det*ksi
return ksi //instead of return newdet/det

Also, a multiscale complex type (complex_as_logphi) is introduced for the value of det and newdet such that even when the value of determinant is underflowing in some MC steps, it can become finite again. complex_as_logphi stores std::abs and std::arg of value_type (double or complex<double>), but abs as log(abs) such that much lower values can be stored than available in double.

### DIFF
--- a/triqs/det_manip/det_manip.cpp
+++ b/triqs/det_manip/det_manip.cpp
@@ -1,0 +1,62 @@
+#include "./det_manip.hpp"
+
+namespace triqs { namespace det_manip {
+
+complex_as_logphi& complex_as_logphi::operator=(std::complex<double> a) {
+  log=std::log(std::abs(a));
+  phi=std::arg(a);
+  return *this;
+}
+
+complex_as_logphi& complex_as_logphi::operator*=(const complex_as_logphi& a) {
+  log+=a.log;
+  phi+=a.phi;
+  return *this;
+}
+
+complex_as_logphi& complex_as_logphi::operator*=(std::complex<double> a) {
+  log+=std::log(std::abs(a));
+  phi+=std::arg(a);
+  return *this;
+}
+
+complex_as_logphi::operator double() const{
+  double a = std::exp(log);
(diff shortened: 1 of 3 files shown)


krivenko:
Thank you for the pull request!

@pseth @parcollet @mferrero 
We should seriously consider this work as a solution to https://github.com/TRIQS/cthyb/issues/5.

parcollet:
Yes we are discussing this with Jaksa and Michel.
Thanks for reminding cthyb/#5

1- the problem arise because of an underflow of the det.

2- one part of the fix is indeed to avoid newdet/det and so on in the try_xxx of the det_manip.
it is useless (we already have the ratio), and lead to nan. This fix is fine.

3- now the second part of the fix is less clear.

->  One option is apparently to simply remove det newdet and determinant() from the class.
We do not use it in the ctqmcs (except in my recent Keldysh algorithm), we only use the *ratios* 
and the user can keep the det. in a variable by multiplying the ratios if needed. 
Then she can use a variable of appropriate type double, long double, or GMP even,
but that is not the business of the det_manip class.

-> other option is to keep det. and newdet with a larger type. But which one ?
a custom class like here is introducing a new type in the lib, without all operations, etc....
GMP already exists after all, with all operations. long double underflow below 10^-4932.
If your det. is < 10^-4932 , probably either it is 0 in the computation  or the calculation has to be rescaled, no ? This is strange ...

opinions ?

parcollet:
Most of it was merged in Jan.

Issue #328, opened 2016-05-22
Can't build with GCC 6


Titan-C:
Maybe is cmake that does not understand the new version of GCC anthough is supported. I get this message 
```
-- Installation directory is /home/oscar/miniconda3/envs/dev2
-- Compiler is gcc with version g++ (GCC) 6.1.1 20160501
Copyright (C) 2016 Free Software Foundation, Inc.
This is free software see the source for copying conditions.  There is NO
warranty not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
CMake Error at cmake/FindCompilerCheck.cmake:79 (MESSAGE):
  

  ************************** FATAL ERROR ************************************

  You are using the gcc compiler but your compiler is too old :

   TRIQS requires version >= 4.8.1 while you have g++ (GCC) 6.1.1 20160501

  Copyright (C) 2016 Free Software Foundation, Inc.

  This is free software see the source for copying conditions.  There is NO

  warranty not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.

    

  ************************** FATAL ERROR ************************************

Call Stack (most recent call first):
  CMakeLists.txt:62 (find_package)


-- Configuring incomplete, errors occurred!

```

parcollet:
yes, it does compile on gcc 6 normally.
this is the simple cmake check ...
what does g++ --version says exactly ?

parcollet:
which distribution by the way  ?

Titan-C:
Olivier Parcollet <notifications@github.com> writes:

> what does g++ --version says exactly ?

oscar@titan-arch-m ~ $g++ --version
g++ (GCC) 6.1.1 20160501
Copyright (C) 2016 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.

> which distribution by the way ?

Is Arch linux

parcollet:
ok, the check of version is not nicely done. I will patch this.

parcollet:
please recheck on branch gcc6, I have pushed a little patch.
Need to recheck on OS X, etc.. before merging.

Titan-C:
Yes it builds.
One question, isn't it simpler to use cmake to check the version instead of using and extra compiled c++ file?
Something like:
```cmake
if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
[...]

PR #339, opened 2016-06-22
Compile under GCC6

Titan-C:
Following up with issue #328 I took the post from branch gcc6 and rebased to current master. It includes for GCC the use of the CMAKE_CXX_COMPILER_VERSION

Tested in archlinux with GCC 6.1.1 and cmake 3.5.2

White-space was cleaned in file


krivenko:
Can't see why this PR has to hang here any longer. Merging.
Thanks for the patch, Óscar!

PR #367, opened 2016-11-24
[atom_diag] New framework, atom_diag

krivenko:
**Lightweight exact diagonalization solver component ported from cthyb.**

C++
---

* I had to add a template parameter to the main `atom_diag` class
  to switch between real and complex matrix types. Cannot really see a
  way around it since different solvers may choose to compile with different
  versions. Of course, the implementation has become a bit more messy but it
  is still manageable. Some macros have been required to reduce typing.
  Main portions of the code are explicitly instantiated in .cpp files so that
  compilation speed is not affected too much.

* `h5_read` will transparently read real-valued data into a complex `atom_diag`
  object.

* Some methods are renamed to have 'subspace' instead of 'block' in their
  names. 'block' was a constant source of confusion when used in one context
  with `block_gf`.

* `act()` and `get_vacuum_state()` assume all many-body states to be written
  in the eigenbasis.

* Renamed function `quantum_number_eigenvalues2()` ->
 `quantum_number_eigenvalues_checked()` to make its purpose clearer.

* Completely reworked `atomic_gf`. It now returns a special object
  (`atomic_block_gf`) containing a list of single-particle transition energies
  and their corresponding residues. This intermediate representation can be used
  to fill any `block_gf<imtime>`, `block_gf<imfreq>` or `block_gf<legendre>` object.
```c++
    auto agf = atomic_gf(ad, beta, gf_struct, excluded_states);

    // Fill block GFs with data
    G_tau() = agf;
    G_iw() = agf;
    G_l() = agf;
```
* Three thorough C++ tests are added (real, complex and h5 operations).

Python
------

* `atom_diag` is wrapped in Python (`AtomDiagReal`/`AtomDiagComplex`) with most of
  its methods.

* All free functions from `atom_diag_functions.hpp` are wrapped.

* `atomic_block_gf` is wrapped such that it can be used as RHS to fill `BlockGf`.
```python
    G_tau << atomic_gf(ad, beta, gf_struct)
    G_iw << atomic_gf(ad, beta, gf_struct)
    G_l << atomic_gf(ad, beta, gf_struct)
```
* Factory function `AtomDiag` returns an instance of `AtomDiagReal`/`AtomDiagComplex`
  depending on the Hamiltonian passed to it.

* Python test is added (atom_diag_python).

Documentation
-------------

* A short introductory paragraph, a couple of use examples
  (C++ & Python) and some cpp2doc-generated documentation.


krivenko:
As pointed out by Michel @mferrero, there is a potential name collision between `atom_diag` namespace and class.
Should we call one of them simply `diag`?
Better ideas anyone?

krivenko:
Possible generalization
------------------------------
- Allow non-abelian quantum numbers in the QN constructor
```c++
atom_diag(many_body_op_t const& h, fundamental_operator_set const& fops, std::vector<many_body_op_t> const& qn_vector);
```

- Modify the auto-partition constructor
```c++
atom_diag(many_body_op_t const& h, fundamental_operator_set const& fops, fundamental_operator_set const& merge_fops = {});
```
Only operators from `merge_fops` will have one-to-one connections.

krivenko:
TODO list for this pull request
-------------------------------------
* [ ] Allow non-abelian quantum numbers.
* [ ] Redesign `atomic_gf()` and remove `atomic_block_gf`.

PR #371, opened 2017-01-11
 added sum_of_elements to minivector

Wentzell:
Necessary for the nfft buffer that is currently beeing added to the ctint / cthyb

### DIFF
--- a/triqs/arrays/algorithms.hpp
+++ b/triqs/arrays/algorithms.hpp
@@ -54,10 +54,12 @@ namespace triqs { namespace arrays {
 
 
  template <class A>
-  typename A::value_type sum(A const & a) { return fold ( std::plus<typename A::value_type>())  (a); }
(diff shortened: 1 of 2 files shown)


krivenko:
Hi Nils,
There is a typo in your commit, `T=0 res`...

Wentzell:
Sorry, fixed.

parcollet:
Ok. Can you rename it to 
friend sum(mini_vector const & ...) ....
the product_of_elements is not a nice name either.
I will partially replace this by std::array in dev.

Wentzell:
I changed it, however I did it with a template function and some forward declarations.
The problem is that if you need to make it visible to the triqs::utility namespace, which is non possible
if you only do it with a non-template member function. 

See also 'Approach 3' in http://web.mst.edu/~nmjxv3/articles/templates.html

Wentzell:
Updated. This is cleaner with a non-template friend, but we need to fix the sum in algorithms.hpp with it.

parcollet:
Merged in dev

PR #375, opened 2017-02-01
[det_manip]

Wentzell:
-Added reject_last_try() function that resets all pending operations
    (i.e. last_try = No_try)
  -API Change: Each try has to be either completed or rejected!
    (to avoid bugs and for code clarity)
  -Adjusted tests accordingly
  -Allow for seperate types of x- (row) and y- (column) index!
    (e.g. to choose different index types for c and cdagger in QMC)

### DIFF
--- a/test/triqs/det_manip/det_manip1.cpp
+++ b/test/triqs/det_manip/det_manip1.cpp
@@ -110,7 +110,10 @@ struct test {
      D.complete_operation();
      if (D.size() >0) check();
     }
-    else { std::cerr  << " reject  since new det is = " << std::abs(detratio*det_old)<<std::endl;}
+    else { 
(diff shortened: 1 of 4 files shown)


HugoStrand:
This breaks cthyb move_global.hpp which assumes that they are of the same type.

Wentzell:
The types are not the problem. 
I only added the possibility for them to be different. They can still be equal. 

The point is that you have now have to either **reject** or **complete** every try_XXX before doing another try_XXX. Rejection is done using
```c++ 
det.reject_last_try() 
```

This of course involves some changes in the CTHyb.

HugoStrand:
Ok, thank you @Wentzell for the guidance. CTHyb/dev is now up to speed on this, see TRIQS/cthyb@d847336ac36575288af606a66e2d3af96a789130.

PR #387, opened 2017-03-30
[atom_diag] New framework, atom_diag (was PR #367)

krivenko:
*This is a follow-up pull request after #367, which was closed by GitHub as `atom_diag` branch had been deleted.*

Recent changes
=============
C++
---

* Removed `atomic_block_gf` class completely.

     Instead, there is a function `atomic_g_lehmann(atom_diag ad, ...)`
     computing an intermediate (Lehmann) representation of the GF.
     The representation itself is a combination of TRIQS arrays and STL containers.

     ```c++
     using gf_scalar_lehmann_t = std::vector<std::pair<double, double/dcomplex>>;
     using gf_lehmann_t = std::vector<matrix<gf_scalar_lehmann_t>>;
     ```

     Once computed, the Lehmann representation can be passed to

        - block_gf<imtime>   atomic_g_tau(gf_lehmann_t lehmann, ...)
        - block_gf<imfreq>   atomic_g_iw(gf_lehmann_t lehmann, ...)
        - block_gf<legendre> atomic_g_l(gf_lehmann_t lehmann, ...)
        - block_gf<refreq>   atomic_g_w(gf_lehmann_t lehmann, ...)

     These four functions have 'high-level' overloads, which take an `atom_diag` object and call `atomic_g_lehmann()` internally.

* Reorganize public headers, now they are three:

    `triqs/atom_diag/atom_diag.hpp`
    `triqs/atom_diag/functions.hpp`
    `triqs/atom_diag/gf.hpp`

* Reproducibility issues of C++ tests have been fixed.

Python
------

Only the 'high-level' functions `atomic_g_tau()`, `atomic_g_iw()`, `atomic_g_l()` and `atomic_g_w()` have been wrapped.


krivenko:
@parcollet @mferrero 
I really need your opinion on the necessity of the intermediate Lehmann representation of GFs (available only in C++ as of my latest commit).

Pro: Computed only once, can be reused many times to calculate `G_tau`, `G_iw`, `G_l`, `G_w` on different meshes, with different energy windows, broadening etc.

Con: Requires memory to store and extra time to (re)allocate storage, even when used only once. I think this is not the limiting factor for applications, because much more memory is required to store C/C^+ matrices.

parcollet:
Merged in unstable.
Possible renaming to atom_diag_abelian or similar to be discussed.

PR #388, opened 2017-03-30
Dev17

Wentzell:
See Commit Messages!


test/triqs/utility/itertools.cpp
```diff
@@ -0,0 +1,128 @@
...
+}
+
+TEST(Itertools, Multi) {
+
+  std::vector<int> V{1, 2, 3, 4, 5, 6};
+
+  // Build enumerate from transform and compare
+  auto l = [n = 0](auto x) mutable { return std::tuple<int, decltype(x)>{n++, x}; };
+
+  for (auto[x1, x2] : zip(transform(V, l), enumerate(V))) {
```
HugoStrand: Beautiful! :)

test/triqs/utility/itertools.cpp
```diff
@@ -0,0 +1,128 @@
...
+};
+
+TEST(Itertools, Zip) {
+
+  std::array<_int, 6> V1{6, 5, 4, 3, 2, 1};
+  std::vector<int> V2{1, 2, 3, 4, 5, 6};
+
+  // Zip both vectors and sum
+  for (auto[x, y] : zip(V1, V2)) {
+    TRIQS_ASSERT(7 - y == x);
```
krivenko: Why do you use `TRIQS_ASSERT` in the test?
It would make more sense to stick to GTest's `EXPECT_TRUE` or `ASSERT_TRUE`.

test/triqs/utility/itertools.cpp
Wentzell: Yes its very nice and compact :)
It looked comparable in speed to explicit writings in a brief test, 
but we need to do some rigorous performance checking this.
In particular if we want to use this technology in performance essential loops

test/triqs/utility/itertools.cpp
Wentzell: That makes a lot of sense! :) 

Thanks for the suggestion, I will update that.

PR #391, opened 2017-04-03
Density() function for GfReFreq

mzingl:
This commit implements the density() function for real frequency Greens function.

- The integration from x_min to 0 is performed with the trapezoidal rule.

- If 0 is not a point of the mesh the integration is carried out up to the last negative mesh point x1.
Additionally, the remaining area Ar from x1 to 0 is added. This area is given by
Ar = f(x1) * |x1| + |x1|^2/(2*delta_w) * (f(x2)-f(x1)), where x2 is the first positive mesh point.

Please feel free to alter or improve this commit. In any case I would really appreciate having a density()
function for GfReFreq on the master branch.

Thanks,
Manuel


Wentzell:
Could you provide a test for this new functionality?

HugoStrand:
@mzingl would it make sense to have a temperature parameter for the density? I.e. to have the option to integrate with a Fermi/Bose function?

HugoStrand:
This has been merged in #534

PR #392, opened 2017-04-06
 Add -march=native to release flags

Wentzell:
This should be always used for any triqs setup to allow for architecture specific optimizations.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -349,6 +349,7 @@ if(NOT CMAKE_BUILD_TYPE)
  mark_as_advanced(CMAKE_BUILD_TYPE)
 endif(NOT CMAKE_BUILD_TYPE)
 message(STATUS "Build type: " ${CMAKE_BUILD_TYPE})
+set(CMAKE_CXX_FLAGS_RELEASE "-O3 -march=native -DNDEBUG" CACHE STRING " The release flags " FORCE)
 
 # Fortran bindings
 option(BIND_FORTRAN_LOWERCASE "FORTRAN functions are compiled WITHOUT a trailing underscore" OFF)


CMakeLists.txt
```diff
@@ -349,6 +349,7 @@ if(NOT CMAKE_BUILD_TYPE)
  mark_as_advanced(CMAKE_BUILD_TYPE)
 endif(NOT CMAKE_BUILD_TYPE)
 message(STATUS "Build type: " ${CMAKE_BUILD_TYPE})
+set(CMAKE_CXX_FLAGS_RELEASE "-O3 -march=native -DNDEBUG" CACHE STRING " The release flags " FORCE)
```
HugoStrand: Good point, but there is a problem with Mac OSX. 

At least for gcc the default system assembler do not support the instructions generated by the compiler. To select the gcc assembler one has to use "-Wa,-q -march=native".

I know this is OSX being broken and we maybe should not care. But this will potentially generate more "can not compile triqs" issue reports.

(Note that "-march=native" is not enough to get AVX-FMA instructions since it changes the floating point round off error.)

Titan-C:
`-march=native` is fine for a local computer. But on an heterogeneous
cluster (one that has grown over the years for example) I guess is
not a good idea(would love to be wrong). Compilation does not occur in
the same computer node as where the code is executed, this different
nodes don't have the same architecture and processor instructions.

Is is possible to have a compiler instruction like they do in OpenBlas
for Dynamic Architecture. That is the library is compiled multiple times
for various architectures and on runtime the correct/most optimized
library for current runtime environment is loaded.

Wentzell:
@HugoStrand Ok, I should have checked on the Mac... I agree that its better to avoid these problems.

@Titan-C Didn't think about that, but it makes a lot of sense. I have no experience with the multi-arch libs.
  Is the usage very straightforward and easy to setup?

I suppose it's better not to default it then and leave it up to the user.

krivenko:
I agree with Óscar @Titan-C.
It was a pain to figure out why one of MPI processes (with a different rank every time!) was crashing with SIGILL on a heterogeneous machine.

PR #396, opened 2017-04-13
[pygf] bugfix pytriqs.gf.local removed

HugoStrand:
Running the tests (dos, brillouin_zone, g_k_om) fails in hilbert_transform.py that tries to import the no longer existing python module pytriqs.gf.local.


parcollet:
Merged already, also fixed doc.

PR #399, opened 2017-04-14
[gf] python double bracket bugfixes + test

HugoStrand:
Found two bugs in the double bracket accessor/slicer and wrote simple test case.


HugoStrand:
@parcollet could you have a brief look at this? I think I found some logical bugs in the double bracket accessor.

PR #402, opened 2017-04-15
[cmake] Properly check <optional> availability (fix #401)

### LINKED ISSUE #401: [unstable] GCC 5.4 cannot compile optional_compat.hpp
As far as I understand `optional_compat.hpp` is a workaround for the older compilers supporting `<experimental/optional>` but not `<optional>`.
```c++
#if __cplusplus > 201402L 
#include <optional>
#else
#include <experimental/optional>
namespace std { // to be found by ADL

 using std::experimental::optional;
 using std::experimental::in_place;
}
#endif
```

Now the problem is that `__cplusplus` is defined to be 201500 on GCC 5.4, while `<optional>` is not yet supported by it.

My compiler is `gcc (Gentoo 5.4.0-r3 p1.3, pie-0.6.5) 5.4.0`.

krivenko:
Commit 90ec42e8a7430beba5161b2acec2f8828fa54826 is too optimistic about compiler's support for `<optional>` header.
Here I check for `<optional>` explicitly.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -230,6 +230,17 @@ else ()
 endif()
 
 # #################################
+# Check availability of <optional>
+# #################################
+try_compile(has_optional_header ${CMAKE_BINARY_DIR} ${CMAKE_SOURCE_DIR}/cmake/has_optional.cpp COMPILE_DEFINITIONS -std=c++1z )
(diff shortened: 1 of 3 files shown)


parcollet:
@Wentzell for review

Wentzell:
Agreed!

PR #403, opened 2017-04-18
Update to FindLibClang.cmake:

Wentzell:
Search for llvm-4.0 and corresponding executables
Add relevant environment variables in Trial paths

@krivenko Do you agree?

### DIFF
--- a/cmake/FindLibClang.cmake
+++ b/cmake/FindLibClang.cmake
@@ -15,13 +15,16 @@ if( ${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
   set(TRIQS_LIBCLANG_CXX_ADDITIONAL_FLAGS "" CACHE STRING "Additional flags to be passed to libclang when parsing with clang")
 
   set(CLANG_COMPILER "clang++")
-  set(CLANG_OPT  -std=c++1y -stdlib=libc++ )
+  set(CLANG_OPT -stdlib=libc++ )
 
 else()
 
 SET(TRIAL_LIBRARY_PATHS
+ ENV LIBRARY_PATH
+ ENV LD_INCLUDE_PATH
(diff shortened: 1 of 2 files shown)


parcollet:
Looks fine.
There is however a real pb on Os X now, one of the include we added 
to libclang cause simple codes to fail to compile ...
Not clear what a clean solution is

cmake/FindLibClang.cmake
```diff
@@ -15,13 +15,16 @@ if( ${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
   set(TRIQS_LIBCLANG_CXX_ADDITIONAL_FLAGS "" CACHE STRING "Additional flags to be passed to libclang when parsing with clang")
 
   set(CLANG_COMPILER "clang++")
-  set(CLANG_OPT  -std=c++1y -stdlib=libc++ )
+  set(CLANG_OPT -stdlib=libc++ )
 
 else()
 
 SET(TRIAL_LIBRARY_PATHS
+ ENV LIBRARY_PATH
```
krivenko: Shouldn't it be `$ENV{LIBRARY_PATH}`?

Wentzell: No, since you have to get rid of the : and put spaces between the different folders in the LIBRARY_PATH.

This is done by cmake automatically, see signature here:

https://cmake.org/cmake/help/v3.0/command/find_library.html

krivenko: Ok, I see.

`ENV` is a keyword in `find_library`'s signature. Are you sure it will be recognized as such when passed via a variable? What happens if I want to search for a library in a directory called `ENV`?

krivenko: Never mind, I checked it and it works.

PR #417, opened 2017-05-23
Corentin conditioning (det_manip)

CorentinB78:
In det_manip, computes the conditioning number at each change. It is then used to estimate the accuracy of the inverse matrix. If the conditioning number is too high a regeneration is called.
The aim is to be able to rely on det_manip to compute inverse when near singular matrices can appear.


parcollet:
After discussion with Alice, Xavier, it appears that this is not a stable, nor optimal solution.
Besides, it slows code by a factor  approx 2.
So we will not merge this as it.
Kept in log of PR, and in our real time QMC codes.

PR #420, opened 2017-05-28
[many_body_operator]

Wentzell:
Added additional constructor to many_body_operator to allow for e.g.

```c++
many_body_operator h_int;
...
many_body_operator h_dens; 
for (auto const &term : h_int)
    if ( is_density_interaction(term) ) h_dens = h_int_dens + term;
```

@krivenko What do you think?
Maybe add a check if the monomial is ordered correctly?

### DIFF
--- a/triqs/operators/many_body_operator.hpp
+++ b/triqs/operators/many_body_operator.hpp
@@ -126,7 +126,7 @@ namespace operators {
 
   struct _cdress; 
   many_body_operator_generic(_cdress const& term) {
-     monomials.insert({term.monomial, term.coef}); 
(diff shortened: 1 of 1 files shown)


krivenko:
Hi Nils,

We should not allow construction of an operator in an invalid state. So yes, a check is necessary.
Another, even better solution is to call `normalize_and_insert()` (see how it's used in `operator*=()`).

Igor

Wentzell:
Hey Igor,

> Another, even better solution is to call normalize_and_insert()

Ah perfect. This should be it. Please confirm before merge

PR #424, opened 2017-06-06
[product-mesh]

Wentzell:
-added index_t constructor taking tuple of indices
-small bugfix in mesh_point
-added member function 'index' for mesh_point<cartesian_product< ... >> to
be consistent with e.g. mesh_point<gf_mesh<imfreq>>


Wentzell:
To be tested on OSX first, seems to lead to problems there.

Wentzell:
@parcollet No Problem with OSX! Can be merged as is!

Issue #364, opened 2016-11-03
Make complex operators real if imaginary part is small


mzingl:
Hi!
When constructing our Hamiltonians we sometimes end up in case where we have complex coefficients which have only a very small imaginary part (< 1e-10). 
To have only real coefficients Gernot wrote a little function in python:
```
def make_operator_real(H):
    H2 = Operator()
    for term in H:
        h = Operator()+1
        for el in term[0]:
            if el[0]:
                h = h*c_dag(*el[1])
            else:
                h = h*c(*el[1])
        if abs(term[1].imag) > 1e-8:
            print 'WARNING: Imaginary part larger than 1e-8'     
        H2 = H2 + term[1].real*h
    return H2  
```
It would be nice to have such a function directly in TRIQS. Maybe it is even possible to automatically convert to real coefficients if the imaginary parts are tiny.

Thanks.

parcollet:
I agree.
Should be a method .real() in the operator class ?
@krivenko : opinion ?

krivenko:
I don't mind adding `real()` and `imag()`, but rather as free functions, just as we did with `dagger()`.
Should it be done on the C++ level?

parcollet:
I guess so, and wrapped.

parcollet:
wrapped as a method

mferrero:
I also think it would be best to avoid an automatic conversion (as proposed by Manuel) to avoid unexpected behaviour. Do you agree?

parcollet:
Sure !
Explicit is better than implicit

mzingl:
My idea to "correct" operators automatically is coming from this:
`print 1e-14*c_dag(1,0)`
which is automatically set to 0. Should this also be explicit?

In the end such terms or terms with small imaginary part are always coming from numerical inaccuracies. Therefore, at least the solver should automatically throw away coefficients which are small and convert coefficients to real if the imaginary part is small.

krivenko:
> My idea to "correct" operators automatically is coming from this:
> `print 1e-14*c_dag(1,0)`
> which is automatically set to 0. Should this also be explicit?

Now, your comment makes me think that only `0*c_dag(1,0)` must be converted to 0...

Anyhow, I shall add `real()`, `imag()` and a pure Python function similar to `make_operator_real`, maybe with an extra tolerance argument.

krivenko:
Here you go!
[...]

PR #430, opened 2017-06-25
[operators] Add friend foreach_monomial()

krivenko:
Following suggestion of @francescopt.

https://github.com/TRIQS/triqs/issues/364#issuecomment-260948562

### DIFF
--- a/triqs/operators/many_body_operator.hpp
+++ b/triqs/operators/many_body_operator.hpp
@@ -287,7 +287,7 @@ namespace operators {
 
   // Transform operator by applying a given function to each monomial
   template <typename Lambda>
-  friend many_body_operator_generic foreach_monomial(many_body_operator_generic const& op, Lambda L) {
+  friend many_body_operator_generic foreach_monomial(many_body_operator_generic const& op, Lambda&& L) {
    many_body_operator_generic res;
    using triqs::utility::is_zero;
    for (auto const& x : op) {
--- a/triqs/operators/many_body_operator.hpp
+++ b/triqs/operators/many_body_operator.hpp
@@ -124,9 +124,9 @@ namespace operators {
    if (!is_zero(x)) monomials.insert({{}, x});
   }
 
-  struct _cdress; 
+  struct _cdress;
   many_body_operator_generic(_cdress const& term) {
-   normalize_and_insert(term.monomial, term.coef, monomials); 
+   normalize_and_insert(term.monomial, term.coef, monomials);
   }
 
   template <typename S> many_body_operator_generic& operator=(many_body_operator_generic<S> const& x) {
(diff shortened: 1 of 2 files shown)


triqs/operators/many_body_operator.hpp
```diff
@@ -285,28 +285,32 @@ namespace operators {
    return res;
   }
 
-  // real
-  friend many_body_operator_generic real(many_body_operator_generic const& op) {
+  // Transform operator by applying a given function to each monomial
+  template <typename Lambda>
+  friend many_body_operator_generic foreach_monomial(many_body_operator_generic const& op, Lambda L) {
```
Wentzell: Pass L as ``` Lambda const&```?

triqs/operators/many_body_operator.hpp
```diff
@@ -285,28 +285,32 @@ namespace operators {
...
    many_body_operator_generic res;
-   using triqs::utility::real;
    using triqs::utility::is_zero;
    for (auto const& x : op) {
-    auto c = real(x.coef);
+    auto c = L(x.monomial, x.coef);
```
Wentzell: What do you have in mind that might change the coeff depending on the monomial as well?

triqs/operators/many_body_operator.hpp
```diff
@@ -285,28 +285,32 @@ namespace operators {
...
     if(!is_zero(c)) res.monomials.insert({x.monomial, c});
    }
    return res;
   }
 
+  // real
+  friend many_body_operator_generic real(many_body_operator_generic const& op) {
+   return foreach_monomial(op, [](monomial_t const&, scalar_t c) {
+    using triqs::utility::real;
+    return real(c);
```
Wentzell: maybe just
```c++
return triqs::utility::real(c);
```

triqs/operators/many_body_operator.hpp
krivenko: Good point. I think a universal reference would make even more sense here.

triqs/operators/many_body_operator.hpp
krivenko: Any sort of filtration algorithm that returns 0 for each unwanted monomial.

triqs/operators/many_body_operator.hpp
krivenko: Compilation of your variant will fail if `triqs::utility::real(scalar_t)` is not defined.
There is still a chance, however, that `real(scalar_t)` is found by argument-dependent lookup, which makes compilation of `real(c)` possible.

krivenko:
O.k., so you want more generic names...

I am fine with `transform`, but `filter` is a special case of `transform` and thus redundant.
These two operations are essentially different for lists: `transform` preserves length of the list, while `filter` does not. For monomials, however, it's equivalent to get a zero coefficient or no coefficient at all.

krivenko:
Just in case, I made the requested change...

francescopt:
Another useful ```transform``` routine would consist in allowing a transformation of the indices. That is, you provide a lambda function ```indices_t->indices_t```, and iterate over the monomials and, for each monomial, over the indices. In principle one can implement such a function with the ```const``` access provided by the iterators, but it should be much more efficient if one introduces a friend function like the ```transform``` outlined above, because then one has direct write access to the monomials and to the ```normalize_and_insert``` routine.

PR #448, opened 2017-07-25
[hdf5] Archive bug casting bool -> numpy.bool_

HugoStrand:
Here is a test that checks that python types comes back with the same type after being mangled in an hdf5archive.

Currently it fails since `bool` is converted to `numpy.bool_`

```
my_flag <type 'bool'> <type 'numpy.bool_'>
  File "hdf5_io_types.py", line 48, in <module>
    assert( type(p[key]) == type(p_ref[key]) )
```

### DIFF
--- a/test/pytriqs/base/CMakeLists.txt
+++ b/test/pytriqs/base/CMakeLists.txt
@@ -11,6 +11,7 @@ add_python_test(operator_python2)
 
 # Some basic HDF IO stuff
 add_python_test(hdf5_io)
+add_python_test(hdf5_io_types)
 
 # Greens functions
 add_python_test(gf_init)
--- a/test/pytriqs/base/hdf5_io_types.py
+++ b/test/pytriqs/base/hdf5_io_types.py
@@ -1,0 +1,55 @@
+################################################################################
+#
+# TRIQS: a Toolbox for Research in Interacting Quantum Systems
+#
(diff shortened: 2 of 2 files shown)


parcollet:
It is a problem of h5py, which we use as a basic layer.
bool, int, float, complex are stored and retrieved as 
numpy._bool, numpy.int64, numpy.float64, numpy.complex128  resp.

A replacement of h5py by a basic cpp2py interface is already scheduled, 
but not before 1.7 or 1.8.
It will solve this pb, along with removing the h5py dependency (which creates other install problems).

Is this an urgent problem ?

HugoStrand:
This is not urgent,

but it is an inconsistency, since the c++2py interface does not support `np._bool` as arguments for functions that take `c++` bools. So if one stores a python bool to a h5 archive and then try to use it as an input parameter to a solver you get a c++ type error.

parcollet:
Problem occurs when reread h5. h5py transform a bool into numpy._bool, 
which wrapped C++ function does not accept.
Quick solution is to fix the converter to accept numpy_bool.
Long term solution is to get rid of h5py.

parcollet:
Do we have a test for this ?

parcollet:
Sorry, my confusion, this was PR, not issue ...

Wentzell:
This test is now passing for triqs/unstable.
We have merged the test into https://github.com/triqs/h5

PR #464, opened 2017-08-16
Change NPY_CHAR to NPY_STRING

Titan-C:
I recompiled my triqs from master and started seeing a lot of deprecation warnings regarding NPY_CHAR. This small patch fixes it.

NPY_CHAR was deprecated in numpy 1.7 but its complete deprecation started
being enforced in numpy 1.13


Wentzell:
We fixed this in unstable already, but its good to have the patch in master as well.

PR #465, opened 2017-08-22
[optional] Fixes #462

### LINKED ISSUE #462: unstable optional_compat.hpp broken for systems with HAS_OPTIONAL_HEADER=true
On mac with clang 4.0 i get compiler errors from `utility/optional_compat.hpp` since I have an `optional.hpp` present on my system.

@Wentzell please help

Wentzell:
Fixes #462
Relates to #401
We have to check for both, the availability of #include <optional> and
the version of c++ (i.e. -std=c++1y vs c++1z)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -232,7 +232,7 @@ endif()
 # #################################
 # Check availability of <optional>
 # #################################
-try_compile(has_optional_header ${CMAKE_BINARY_DIR} ${CMAKE_SOURCE_DIR}/cmake/has_optional.cpp COMPILE_DEFINITIONS -std=c++1y )
+try_compile(has_optional_header ${CMAKE_BINARY_DIR} ${CMAKE_SOURCE_DIR}/cmake/has_optional.cpp COMPILE_DEFINITIONS ${TRIQS_CXX_DEFINITIONS} )
 if (NOT has_optional_header)
  message(STATUS "Using <experimental/optional> header instead of <optional>")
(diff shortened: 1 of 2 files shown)


krivenko:
What's the actual problem here?
In what situation does `has_optional.cpp` compile whereas triqs itself does not? (Issue #462  lacks any details...)

Wentzell:
The problem is that even if you ```#include <optional>``` it will not have ```optional``` in the ```std``` namespace unless you pass ```-std=c++1z```. This is checked by __cplusplus > 201402L as in https://github.com/gcc-mirror/gcc/blob/master/libstdc%2B%2B-v3/include/bits/c%2B%2B17_warning.h

Maybe it would be cleaner though to change the has_optional.hpp to 

```c++
#include <optional>
int main() { std::optional<int> a; return 0; }
```

@krivenko What do you think?

krivenko:
> Maybe it would be cleaner though to change the has_optional.hpp to
>
>```c++
>#include <optional>
>int main() { std::optional<int> a; return 0; }
>```

This is what I thought.
`has_optional.hpp` must compile only if it's found and actually usable.

There is another thing, which is maybe worth changing.

https://github.com/TRIQS/triqs/blob/4986646ad7f0791bac41b5c99ebc969d620ea95b/CMakeLists.txt#L235

I guess putting `${TRIQS_CXX_DEFINITIONS}` instead of `-std=c++1y` makes more sense here.

PR #444, opened 2017-07-20
((WIP do not merge)) [clef] products of matsubara frequencies

HugoStrand:
Clef expressions with multiplication of Matsubara frequency placeholders, e.g.

```c++
g2(om_, nu_)(i_, j_) << i_ + j_ + om_ * (nu_ + om_);
```

does not compile, giving the error:

```
/triqs/triqs/./clef/adapters/../clef.hpp:199:35: error: invalid operands to binary
      expression ('const triqs::gfs::mesh_point<gf_mesh<triqs::gfs::imfreq> >' and 'const
      triqs::gfs::mesh_point<gf_mesh<triqs::gfs::imfreq> >')
 TRIQS_CLEF_OPERATION(multiplies, *);
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~
```

Here are two tests that attempts to do frequency products using placeholders. @parcollet is this something fixable?

It is useful for e.g. computing the atomic vertex of the Hubbard-atom.

Best, H

PR #470, opened 2017-08-29
[matsubara] Product of matsubaras, additional gf product test

Wentzell:
-Implemented product of two matsubara frequencies to yield a complex
-Added a test that calculates and checks the product of two gfs

### DIFF
--- a/test/triqs/gfs/gf_prod.cpp
+++ b/test/triqs/gfs/gf_prod.cpp
@@ -1,0 +1,78 @@
+#include <triqs/test_tools/gfs.hpp>
+#include <triqs/gfs.hpp>
+
+using namespace triqs::clef;
+using namespace triqs::lattice;
+
+TEST(Gf, ProdGfTail) {
+
+  double beta = 1;
+  placeholder<0> iw_; 
+
+  auto G = gf<imfreq, matrix_valued>{{beta, Fermion, 50}, {2,2}};
+  G(iw_) << 1.0 / iw_ + 0.25 / ( iw_ * iw_ * iw_ ); 
+
+  auto GG = G; 
+
+  std::cout << " Tail of G \n " << GG.singularity() << "\n"; 
+
+  GG *= G; 
+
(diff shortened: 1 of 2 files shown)


parcollet:
Great.
Should we at some point (not now) get rid of this CRTP for the mesh point operators ?

parcollet:
Can you merge the test of #444 with this ?

Wentzell:
Yeah, maybe we can just put all the operators directly into matsubara? The only special cases are
addition and subtraction right? In every other case we expect a cast to dcomplex.

Wentzell:
I merged the tests of #444 into gf_prod.hpp

However, looking at the output of the test, it seems that the tails are not set at all?
```
[==========] Running 3 tests from 1 test case.
[----------] Global test environment set-up.
[----------] 3 tests from Gf
[ RUN      ] Gf.ProdGfTail
 Tail of G 
 empty tail

 Tail of product G*G 
 empty tail

 Expected Tail 
 empty tail

[       OK ] Gf.ProdGfTail (6 ms)
[ RUN      ] Gf.AutoAssignMatrixGf2_product
[       OK ] Gf.AutoAssignMatrixGf2_product (1 ms)
[ RUN      ] Gf.AutoAssignMatrixGf2_sum_product
[       OK ] Gf.AutoAssignMatrixGf2_sum_product (1 ms)
[----------] 3 tests from Gf (8 ms total)

[----------] Global test environment tear-down
[==========] 3 tests from 1 test case ran. (8 ms total)
[  PASSED  ] 3 tests.
```
How can this be?

parcollet:
Because this is a serious bug in the tail !
t = omega + 0 + 0/omega + ....
t*t =  omega ^2 ...
t*t*t = omega ^ 3 ? 
BUT the _tail type is truncated at omega^2, so the omega^3 is dropped, so 
t*t*t = 0 !, 
which inverse is undefined, hence "empty tail" or nan...

The tail nightmare again : 
@mferrero : since you are in charge of the tails, what do we do ?

parcollet:
Ok, the issue with tail is moved to an issue.

PR #482, opened 2017-09-14
Tailfit complex fix + improvements (master)

gkraberger:
Qualitatively, this is the same pull request as #481 but based on the master branch.

- This fixes a bug in the tail fitting routines for complex Green's function objects. In that case, two windows have to be given (already now), one for positive and one for negative Matsubara frequencies. Currently, the fit coefficients of the two windows are averaged, which is incorrect and leads to physically wrong tail fits. In this fix, the two windows are fitted in one go.
- Typically, the error of the self-energy roughly goes like omega_n^2, however during the fitting a constant error is assumed. This introduces a parameter `omega_error` that is the exponent of the Matsubara frequency-dependence of the error; setting `omega_error=2` gives the quadratic behavior mentioned. The default is `omega_error = 0`, i.e. the current behavior with constant error.

### DIFF
--- a/pytriqs/gf/local/gf_desc.py
+++ b/pytriqs/gf/local/gf_desc.py
@@ -454,13 +454,13 @@ for py_type, has_tail in [("GfImFreq", True) ,("GfImFreqNoTail",False)]:
               doc = """Fills self with the legendre transform of gl""")
 
   g.add_method(name = "fit_tail",
-              signature = "void(tail_view known_moments, int max_moment, int n_min, int n_max, bool replace_by_fit = true)",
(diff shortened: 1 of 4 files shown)


mzingl:
Hi,
what is the status of this pull request?
On the current master version the real tail_fit is always called, which throws an error as soon as the self energy in tau is complex...
Would be good to review/pull this request.

Thanks,
Manuel

parcollet:
This is now obsolete with the new master, where tails have been completely reworked.
Some parts of the commit have been used in 2.0 
The weighted fit could be reimplemented in 2.0.
@Wentzell.
Thanks,

PR #485, opened 2017-09-19
[brew] Changes necessary for brew formula

Wentzell:
I am working on incorporating a triqs formula into homebrew/science 
(See https://github.com/Homebrew/homebrew-science/pull/5931)

For this some minor changes in the CMake process are necessary to allow for compatibility with the system python on MacOS. Since it is installed in an unconventional path (not containing 'lib') the previous ```REGEX REPLACE ".*/lib"``` was not working properly. 

Further, since homebrew will set up a python virtual-env with the required python packages (as required for homebrew formulas with additional python dependencies), we have to adjust the PYTHONPATH in the pytriqs, ipytriqs and ipytriqs_notebook wrappers.


parcollet:
Shouldn't we remove pythias, ipytriqs, and co ?

Wentzell:
I agree it would be best to replace these wrappers by a simple ```triqs-env.sh``` file that you source if you want to add triqs to your environment. But maybe not on master? This would be a change for a new major version that we need to document well in the changelog.

parcollet:
ah no, sure, not on master...

Wentzell:
@parcollet Can you approve the changes? Or comment on them?

PR #490, opened 2017-10-03
[gf,many_body_operator] Small fixes and additions

Wentzell:
**[gf,block_gf]** Added type real_t to gfs and all target types
    -gf, block_gf, block2_gf now have an interal real_t that corresponds to
    the real version of the Green function type    
    -made the is_gf_real function slightly more restrictive
    -The different target types that can be used for a gf now all have an
    internal real type
    -**BugFix**: __make_zero( .. ) in gf needs special behaviour also for scalar_real_valued

**[many_body_operator]** Small fix and operator==
    -added == operator for many_body_operator_generic
    -normalize_and_insert should take monomial by copy

### DIFF
--- a/triqs/operators/many_body_operator.hpp
+++ b/triqs/operators/many_body_operator.hpp
@@ -383,7 +383,7 @@ namespace operators {
  // Assert that two many-body-operators are term-wise close
  template<typename scalar1_t, typename scalar2_t>
  void assert_operators_are_close(many_body_operator_generic<scalar1_t> const &op1, many_body_operator_generic<scalar2_t> const &op2, double precision){ 
-  auto terms_are_equal = [precision]( auto const &t1, auto const &t2 ){ using std::abs; return (t1.monomial == t2.monomial) && abs(t1.coef - t2.coef) < precision; }; 
+  auto terms_are_equal = [precision]( auto const &t1, auto const &t2 ){ return (t1.monomial == t2.monomial) && triqs::utility::is_zero(t1.coef - t2.coef, precision); }; 
   if(op1.get_monomials().size() != op2.get_monomials().size())
     TRIQS_RUNTIME_ERROR << " ASSERTION FAILED: Operators have different number of terms"; 
   if(!std::equal(op1.begin(), op1.end(), op2.begin(), terms_are_equal))
--- a/triqs/operators/many_body_operator.hpp
+++ b/triqs/operators/many_body_operator.hpp
@@ -23,6 +23,7 @@
 
 #include <ostream>
 #include <cmath>
+#include <algorithm>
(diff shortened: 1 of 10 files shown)


Wentzell:
After discussion with @parcollet I replaced the operator== with a corresponding
assert_operators_are_close function alongside the proper test-macros

triqs/operators/many_body_operator.hpp
```diff
@@ -379,6 +380,16 @@ namespace operators {
   }
  };
 
+ // Assert that two many-body-operators are term-wise close
+ template<typename scalar1_t, typename scalar2_t>
+ void assert_operators_are_close(many_body_operator_generic<scalar1_t> const &op1, many_body_operator_generic<scalar2_t> const &op2, double precision){ 
+  auto terms_are_equal = [precision]( auto const &t1, auto const &t2 ){ using std::abs; return (t1.monomial == t2.monomial) && abs(t1.coef - t2.coef) < precision; };
```
krivenko: For the comparison, I would stick to the pattern used in the rest of the class:
```c++
auto terms_are_equal = [precision]( auto const &t1, auto const &t2 ){
  using triqs::utility::is_zero;
  return (t1.monomial == t2.monomial) && is_zero(t1.coef - t2.coef, precision);
};
```

Unfortunately, my implementation of `is_zero()` was a bit sloppy, so I also suggest updating `/triqs/utility/numeric_ops.hpp`:
```c++
//
// Zero value tests
//
template<typename T> // Integral types
std14::enable_if_t<std::is_integral<T>::value,bool> is_zero(T const& x, T = 0) {
 return x==0;
}

template<typename T> // Floating-point types
std14::enable_if_t<std::is_floating_point<T>::value,bool> is_zero(T const& x, T tolerance = 100*std::numeric_limits<T>::epsilon()) {
 return std::abs(x) < tolerance;
}

template<typename T> // std::complex
std14::enable_if_t<triqs::is_complex<T>::value,bool> is_zero(T const& x, T tolerance = 100*std::numeric_limits<typename T::value_type>::epsilon()) {
 return is_zero(std::real(x), std::real(tolerance)) && is_zero(std::imag(x), std::real(tolerance));
}
```

krivenko:
@Wentzell I'm not too happy about it either.
Removing `int tol` will break `terms_are_equal` for `ScalarType = int`. On the other hand, who is going to use `many_body_operator_generic<int>` anyway...
So it's basically up to you what to do with that argument.

krivenko:
@Wentzell The more I think about the issue, the more I convince myself that you are right.
Even if someone wants to instantiate `many_body_operator_generic<int>`, calling `assert_operators_are_close()` for such an object makes really no sense, so we could simply disallow doing that.

PR #491, opened 2017-10-04
[gf]  mesh_point in Python

parcollet:
Open for discussion : 

- wrap the mesh_point of C++ into MeshPoint in Python

- MeshPoint contains :
   - linear_index
   - value : the result of the C++ cast_t operator (complex, double, k point, etc...).

- if g is Gf
   g.mesh is an iterator which now returns MeshPoint.
   g.mesh.values() is the previous behaviour of g.mesh:
   equivalent to [x.value for x in g.mesh]

- API BREAK :
  in all previous codes, you may have to replace
   g.mesh --> g.mesh.values()
  Painful, but I do not see how to escape this ...

- "may" since  the MeshPoint has arithmetic, float, complex conversion, as in C++.
  This allows to write :
 ``` g= GfImfreq(...)
  for w in g.mesh:
    g[w] = 1/(w + 2) # beware it does NOT set the tail !
```
  
 ```
 g = Gf (mesh = MeshProduct(BZ, W) ....)
  for k, w in g.mesh :
    g[k, w] = expression(k,w)
```
  
```d = [exp(x) for x in g.mesh]```
 will work, as the function will convert to float or complex.

- MeshProduct has a  [ ]
  G.mesh[0]  : same are G.mesh.components[0]

- e.g.

Old example :
```  
   for (i1, t1), (i2, t2), (i3, t3) in itertools.product(*[ enumerate(mesh) for mesh in g4_tau.mesh.components]):
     t1, t2, t3 = t1.real, t2.real, t3.real
     g4_tau[[i1, i2, i3]][:] = g_tau(t1-t2)*g_tau(t3) - g_tau(t1)*g_tau(t3-t2)
```
New writing :
 ``` 
     for t1, t2, t3 in g4_tau.mesh:
      g4_tau[t1, t2, t3] = g_tau(t1-t2)*g_tau(t3) - g_tau(t1)*g_tau(t3-t2)
```
- g[args...]

 If args is only MeshPoints, it is a mesh access
 If args contains at least one MeshPoint and slices, it is interpreted as a slice of the Gf on the mesh. (replace g[[ ]] writing).
 Else it is a target slice as before.

e.g.
 ```
G = Gf (mesh = MeshProduct(BZ, W) ....)
 for k in G.mesh[0] :
    myslice = G[k,:] # a GfImfreq on W only
```

### CODE REVIEW

**Comment by parcollet** [comment]
Do not merge yet, I am opening this for discussion, and to finalise it.

**Comment by parcollet** [comment]
@mferrero, @HugoStrand, @Wentzell : opinions ?

**Comment by HugoStrand** [comment]
I am very happy to have this cleaned up! 👍 

Are all meshpoints treated equally and fairly? I.e. the expressions are the same for MeshImFreq, MeshImTime, etc. ?

Side comment: Currently MeshImTime has the quirk that it returns complex typed times while the greens function only takes real arguments, so one has to "cast to real".

```python
for t in g_tau.mesh:
    print g_tau(t.real) # breaks without .real
```

parcollet:
- yes, all mesh_point are the same (of course, their .value is of different types).
- MeshImTime : the commit fixes this too, .value is the C++ cast_t, so double in this case.
>Are all meshpoints treated equally and fairly? I.e. the expressions are the same for MeshImFreq, >MeshImTime, etc. ?
what do you mean by "expression are the same" ? 

the technique is the same as in C++ : 
g[w] = 1/(w + 2) 
w is a mesh point, but to add it to 2, it is cast into w.value, etc...

parcollet:
PS : there is only one MeshPoint, it is polymorphic

HugoStrand:
Very good.

I will fix `pyed` once you push this to `unstable`, it will clean up the code nicely. :)

krivenko:
Thank you, @parcollet! This is a very nice usability improvement!

> linear_index

Does it run from `0` to `len(g.mesh)-1`?
If so, I'd like to see a way to access indices of Matsubara frequencies, i.e. 0 for the first positive fermionic Matsubara.

Wentzell:
Merged in unstable

PR #505, opened 2017-11-08
[doc] triqs includes to c++2rst

HugoStrand:
`c++2rst` needs the include paths for `triqs`, here is one solution to get them from the `triqs` target.

Please merge

### DIFF
--- a/doc/CMakeLists.txt
+++ b/doc/CMakeLists.txt
@@ -33,6 +33,9 @@ set(CPP_NAMESPACES -N triqs::det_manip -N triqs::mc_tools)
 
 # get include paths for the triqs target
 get_property(TRIQS_INCLUDE_DIRS TARGET triqs PROPERTY INCLUDE_DIRECTORIES)
+foreach(I ${TRIQS_INCLUDE_DIRS})
+ set (INC  ${INC} -I ${I})
+endforeach()
 
 add_custom_command(OUTPUT ${CMAKE_CURRENT_BINARY_DIR}/cpp2rst.log DEPENDS triqs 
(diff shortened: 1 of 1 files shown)


parcollet:
Ok, that is the idea.
@Wentzell 

INCLUDE_DIRECTORIES is the *public* or *private* flags ?
Isn't rather INTERFACE_INCLUDE_DIRECTORIES
?
Cf 
https://cmake.org/cmake/help/latest/command/target_include_directories.html#command:target_include_directories

Wentzell:
INCLUDE_DIRECTORIES contains both the public and the private directories.

The INTERFACE_INCLUDE_DIRECTORIES could be used by the applications using triqs and
would contain the headers in the **installed** location (but it is currently).
Note however that the usage of INTERFACE_INCLUDE_DIRECTORIES is not necessary since
triqsvars.sh is already setting the proper CPATH.

Either way INTERFACE_INCLUDE_DIRECTORIES could not be used before the actual installation.

HugoStrand:
```
get_property(TRIQS_INCLUDE_DIRS TARGET triqs PROPERTY INCLUDE_DIRECTORIES) message(STATUS "TRIQS_INCLUDE_DIRS = ${TRIQS_INCLUDE_DIRS}") string (REPLACE ";" "\n--includes " TRIQS_INCLUDE_DIRS_STR "${TRIQS_INCLUDE_DIRS}") set( TRIQS_INCLUDE_DIRS_STR "--includes ${TRIQS_INCLUDE_DIRS_STR}" )
```

Wentzell:
Already fixed in the current unstable branch

PR #506, opened 2017-11-09
Install modulefile for the module command

krivenko:
Install a TCL modulefile similar to `triqsvars.sh`. It can be used for easy integration with the module systems on HPC machines.

*N.B.* Now that applications are allowed to be installed under different prefixes, they could provide their own modulefiles.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -279,3 +279,9 @@ MESSAGE(STATUS "*   source ${CMAKE_INSTALL_PREFIX}/share/triqsvars.sh          "
 MESSAGE(STATUS "*                                                              ") 
 MESSAGE(STATUS "*   to set up the environment variables                        ") 
 MESSAGE(STATUS "***************************************************************")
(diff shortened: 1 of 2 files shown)


parcollet:
Great.
Is it the file needed for the "module" command ?
Indeed, we will need one per repository : 
- cpp2y (which must be loaded in triqs module, right ?)
- cthyb, ctint, and co.

krivenko:
>  Is it the file needed for the "module" command ?

Yes.

> cpp2y (which must be loaded in triqs module, right ?)

Oh, that's right.
The triqs module must then be updated to include `prereq cpp2py` instruction.

parcollet:
Ok, to be updated then.

Wentzell:
merged, cf. 1b8eae29d92037b64fcd1ac9460c613a8c8f6aa4 and b997c3c6470d0b38b16d4ff37fdad2f6f07247cf

PR #510, opened 2017-11-16
Fixed bug: MPI_C_COMPILE_FLAGS must be a list

shinaoka:
In some environment where MPI_COMPILE_FLAGS is set to " -fexceptions  -fstack-protector-strong  -fPIC",
this string is passed to target_compile_options().
But target_compile_options() takes a list not a string.
This PR is for fixing this problem.


Wentzell:
Fixed in 45c7dea23e3051ecb89ab978a358d10639f4e2ef

PR #511, opened 2017-11-16
[mpi] Fix #508

### LINKED ISSUE #508: [unstable][arrays] matrix transpose() + broadcast()
The following test fails for me when run on >1 cores.
```c++
TEST(Arrays, transpose_bcast) {

 mpi::communicator world;

 matrix<dcomplex> A = {{1, 2, 3},
                       {4, 5, 6}};
 matrix<dcomplex> At = A.transpose();

 matrix<dcomplex> B;
 if(world.rank() == 0) B = At;

 mpi_broadcast(B, world, 0);

 EXPECT_ARRAY_EQ(At, B);
}
```
(`B` ends up having the same shape as `At`, but different traversal order).

This problem apparently has something to do with  #451.

krivenko:
Don't know if I chose the right place for that `set_mem_layout_or_check_if_view()`  function...

### DIFF
--- a/triqs/arrays/impl/indexmap_storage_pair.hpp
+++ b/triqs/arrays/impl/indexmap_storage_pair.hpp
@@ -96,10 +96,10 @@ namespace triqs { namespace arrays {
     static constexpr int rank = IndexMapType::domain_type::rank;
     static constexpr bool is_const = IsConst;
 
-    static std::string hdf5_scheme() { 
+    static std::string hdf5_scheme() {
        return "array<" + triqs::h5::get_hdf5_scheme<value_type>() + "," + std::to_string(rank) + ">";
     }
-    
+
    protected:
 
     indexmap_type indexmap_;
@@ -170,7 +170,6 @@ namespace triqs { namespace arrays {
 
     auto const & memory_layout() const { return indexmap_.memory_layout(); }
     indexmap_type const & indexmap() const {return indexmap_;}
-    indexmap_type & indexmap() {return indexmap_;}
     storage_type const & storage() const {return storage_;}
     storage_type & storage() {return storage_;}
 
@@ -343,6 +342,13 @@ namespace triqs { namespace arrays {
       storage_ = StorageType(indexmap_.domain().number_of_elements());
     }
 
+    void resize (domain_type const & d, memory_layout_t<rank> const & ml) {
+     indexmap_ = IndexMapType(d, ml);
+     // optimisation. Construct a storage only if the new index is not compatible (size mismatch).
+     if (storage_.size() != indexmap_.domain().number_of_elements())
+      storage_ = StorageType(indexmap_.domain().number_of_elements());
+    }
+
     template<typename Xtype>
      void resize_and_clone_data( Xtype const & X) { indexmap_ = X.indexmap(); storage_ = X.storage().clone(); }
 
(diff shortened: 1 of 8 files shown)


triqs/arrays/impl/indexmap_storage_pair.hpp
```diff
@@ -170,6 +170,7 @@ namespace triqs { namespace arrays {
 
     auto const & memory_layout() const { return indexmap_.memory_layout(); }
     indexmap_type const & indexmap() const {return indexmap_;}
+    indexmap_type & indexmap() {return indexmap_;}
```
Wentzell: I don't think its a good idea to expose the indexmap via a non-const reference.
The few functions that need actually change indexmap_ should be friend functions.

triqs/arrays/mpi.hpp
```diff
@@ -70,8 +70,13 @@ namespace arrays {
   template <typename A> REQUIRES_IS_ARRAY mpi_broadcast(A &a, mpi::communicator c = {}, int root = 0) {
    if (!has_contiguous_data(a)) TRIQS_RUNTIME_ERROR << "Non contiguous view in mpi_broadcast";
    auto sh = a.shape();
+   auto m_pos = a.indexmap().memory_layout().get_memory_positions();
    MPI_Bcast(&sh[0], sh.size(), mpi::mpi_datatype<typename decltype(sh)::value_type>(), root, c.get());
-   if (c.rank() != root) resize_or_check_if_view(a, sh);
+   MPI_Bcast(&m_pos[0], m_pos.size(), mpi::mpi_datatype<typename decltype(m_pos)::value_type>(), root, c.get());
+   if (c.rank() != root) {
+     resize_or_check_if_view(a, sh);
+     set_mem_layout_or_check_if_view(a, memory_layout_t<A::rank>(m_pos));
```
Wentzell: This function should NEVER be invoked without a proper resize, am I correct?
Maybe we could just wrap the two operations into one function.

triqs/arrays/impl/indexmap_storage_pair.hpp
krivenko: It is a bad idea, I agree, but how would you implement `set_mem_layout_or_check_if_view()` then?
Making it a friend of `array`, `matrix` and `vector` isn't a great solution either.

triqs/arrays/mpi.hpp
krivenko: Do you want to call `resize_or_check_if_view()` from `set_mem_layout_or_check_if_view()` or rather call both of them from yet another function? Can you suggest a name for the combined function?

Wentzell:
How about the following solution:

- Adjust the resize functions to take a memory layout defaulted to the one of the amv
- Adjust the ```resize_or_check_if_view``` with a default accordingly
- Get rid of set_mem_layout_or_check_if_view
- Keep only ```const &``` accessor of indexmap

If we want to possibility to ever change the memory layout after construction of an amv,
it makes sense to put it into resize in my opinion.

@parcollet What do you think?

parcollet:
I agree with Nils.
1- indexmap should be immutable.
2- Indeed the layout must be broadcasted, so the best solution indeed seems to have
a resize overload taking both the sizes and the layout.

krivenko:
I am trying to rewrite the fix according to your suggestions.
It is a bit difficult to overload `array & array::resize (Args && ... args)`...

parcollet:
ok, fine.
Thanks !

PR #521, opened 2017-12-08
[pade] Fix a problem that arises for large n_points

j-otsuki:
The present Pade code returns NaN when number of Matsubara is large. This pull request fixes this problem. What is changed is that variables A1, A2, B1, B2 are normalized in every step of the iteration so that those do not overflow. It avoids NaN and keeps the ratio A2/B2 (final result).

You can reproduce the above problem using the sample code in
https://triqs.ipht.cnrs.fr/1.x/reference/gfs/py/block/GfReFreq.html
Increase L from 101 to, e.g. 1001, and you will obtain NaN if you print g_pade.data.
In the fixed version, you can obtain correct results for arbitrary number of L.

Best,
Junya Otsuki

### DIFF
--- a/triqs/utility/pade_approximants.hpp
+++ b/triqs/utility/pade_approximants.hpp
@@ -108,8 +108,8 @@ class pade_approximant {
    for(int i=0; i<=N-2; ++i){
      dcomplex Anew = A2 + (e - z_in(i))*a(i+1)*A1;
      dcomplex Bnew = B2 + (e - z_in(i))*a(i+1)*B1;
-     A1 = A2; A2 = Anew;
-     B1 = B2; B2 = Bnew;
+     A1 = A2/Bnew; A2 = Anew/Bnew;
+     B1 = B2/Bnew; B2 = 1.0;
    }
 
    return A2/B2;


Wentzell:
Your change does not give the same result in my opinion.
Could you add a check that tests this?

If we look at the old code we solve for the following iteratively defined sequence at a given index N
```
a(n) = a(n-1) + c(n) * a(n-2)
b(n) = b(n-1) + c(n) * b(n-2)
```

You replace this by a new sequence

```
ap(n) = ap(n-1)/bp(n-1) + c(n) * ap(n-2) / bp(n-1)
bp(n) = 1.0 + c(n) * bp(n-2) / bp(n-1)
```

In order for this to yield the same ratio a(N)/b(N) = ap(N)/bp(N) we require that 

```
ap(n) = f(n) * a(n)
bp(n) = f(n) * b(n)
```

for some complex f(n), which does not seem to work out in your example. 

Maybe you inteded to use

```
f(n) = 1.0 / prod(bp(i), i=0 .. n-1)
```

which would mean

```
          ap(n) = 1.0 / prod(bp(i), i=0 .. n-1) * ( a(n-1) + c(n) * a(n-2) )
<=>   ap(n) =  a(n-1) / (prod(bp(i), i=0 .. n-2) * b(n-1)) + c(n) * a(n-2) / (prod(bp(i), i=0 .. n-3) * bp(n-2) * bp(n-1) )
<=>   ap(n) =  ap(n-1)/b(n-1) + c(n) * ap(n-2)/bp(n-2)/bp(n-1) )
```

while for bp

```
bp(n) =  ( 1.0 + c(n) / b(n-1) )
```

This should indeed normalize the coefficients and fix the problem

j-otsuki:
What is done in the modified code is not that complicated.
Let me explain briefly.

First, look at the following code, that is, the original loop plus additional two lines in the end: 
```
   for(int i=0; i<=N-2; ++i){
     dcomplex Anew = A2 + (e - z_in(i))*a(i+1)*A1;
     dcomplex Bnew = B2 + (e - z_in(i))*a(i+1)*B1;
     A1 = A2; A2 = Anew;
     B1 = B2; B2 = Bnew;

     dcomplex c = arbitrary_value;
     A1 /= c; A2 /= c; B1 /= c; B2 /= c
   }
```
The last two lines in the loop do not change the final result A2/B2 (in analytical sense).
But numerically, the result could be different: One can avoid overflow/underflow by properly choosing the value of c.
An easy choice is 
```
     dcomplex c = B2;
```
With this choice, the above loop may be simplified as
```
   for(int i=0; i<=N-2; ++i){
     dcomplex Anew = A2 + (e - z_in(i))*a(i+1)*A1;
     dcomplex Bnew = B2 + (e - z_in(i))*a(i+1)*B1;
     A1 = A2/Bnew; A2 = Anew/Bnew;
     B1 = B2/Bnew; B2 = 1.0;
   }
```
This is what I've done in the PR.

Could you review again?

Wentzell:
You are correct. My argument was flawed because you are actually changing previous values in the series with your division as well. 
This will indeed maintain the ratio after every loop.

I will port this change to unstable as well.

j-otsuki:
Thank you for your review and merging.

PR #531, opened 2018-01-26
[gf/tail] Bugfix: NAN propagation in tail

Wentzell:
To get proper results from tail (and GF) multiplication, NAN from the first
operand should not propagate if the component of second operand vanishes!

@parcollet Please review and merge!

### DIFF
--- a/triqs/gfs/singularity/tail.hxx
+++ b/triqs/gfs/singularity/tail.hxx
@@ -1170,7 +1170,11 @@ namespace triqs {
         // hence p <= min ( a.n_max, n-b.n_min ) and p >= max ( a.n_min, n- b.n_max)
         const int pmin = std::max(l.order_min(), n - r.order_max());
         const int pmax = std::min(l.order_max(), n - r.order_min());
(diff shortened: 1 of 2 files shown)


parcollet:
I do not get it : NaN * 0 is still NaN, it is normal no ?

mferrero:
In the tails, our understanding of NaN is “unknown”. But we nevertheless know that the number that should be in place of a NaN is a finite number (it cannot be infinite). In that sense 0 * Nan or really 0 * “unknown” is 0 and not NaN. @Wentzell Did I get this right?

Wentzell:
@mferrero Exactly. 

@parcollet Example: Multiplying the tail [0, 0, 0, 1, Nan, ... ] (i.e. 1/w + O(1/w^2)) by itself should result in (0, 0, 0, 0, 1, Nan, ... ) (i.e. 1/w^2 + O(1/w^3)). Currently the result is (0, 0, Nan, ... )

```python
from pytriqs.gf import TailGf
tail = TailGf((1,1),3,-2)
tail[1] = [[1.]]
print tail
print tail * tail
```

Wentzell:
merging after agreement with @parcollet and @mferrero

PR #534, opened 2018-02-05
[gf] GfReFreq density density_zero_T + test

HugoStrand:
This is a remake of pull-request #391 with added text and finite temperature functionality. 

Contains .density(beta) and .density_zero_T() functions for GfReFreq.


mzingl:
To
1.) My opinion is that density() should give the density at T=0. So the default density for GfReFreq should be without Fermi function.
2.) Removing it will brake backward compatibility, or? I guess many are using it in DMFT scripts, at least I do.

HugoStrand:
I have tried to rebase this and fix the comments. But, since I wrote this in February... the whole structure of the python bindings of the greens function has changed.

The files `gf_desc.py`, `functions.cpp` and `functions.hpp` do no longer exist, and the structure in the python bindings seems no longer to add these functions as methods to the Greens function class but rather as free functions.

@Wentzell how and where should the wrapping be done? I am not keen on adding this to `gf_fnt.py` see #635.

Best, H

parcollet:
Ok, we will update the wrappers.

HugoStrand:
I have rebased on unstable, moving the implementation to `gfs/functions/density.*` and put the wrappers in `gf_fnt.py` and removed the `zero_T` methods.

To use the suggested overloading where `G.density()` is zero temperature and `G.density(beta)` is finite temperature I had to add `*args` to the `density` and `total_density` methods in `gf.py` and `block_gf.py`.

With that I consider all comments accounted for.

HugoStrand:
This is pending code review. From @Wentzell or @parcollet .

HugoStrand:
@parcollet this passed the CI, so I merge it.

PR #544, opened 2018-02-28
Improved pade approximants for numerically orbital diagonal data

shinaoka:
Hi, here is a further improvement from one of our colleagues.

When we continue numerically orbital diagonal data, we may encounter "division by zero".
This patch fixes this problem by truncating the expansion in such cases.


shinaoka:
Something went wrong with travis. But, the patch itself looks OK.

shinaoka:
Hi, this is a small fix.
Could you review this patch?

parcollet:
@mferrero : Can you review this ? 
Pade is your contribution...

shinaoka:
@mferrero 
Hello, do you need any help from us?

mferrero:
Sorry for the delay. I merged the request on master. I will do the same on the unstable branch.

PR #550, opened 2018-03-10
Bad bug in TightBinding, checking hopping matrix dims, not active

HugoStrand:
I spent a few hours trying to understand why my dmft loop crashed about 20% of the times the script was run.

It turns out I was passing hopping matrices with incorrect dimensions to `TightBinding` and for some reason it is accepted without any complaints. This causes the lattice Green's function computed by `Sumk` to have uninitialized data, hence the intermittent crashing behavior.

Someone was proactive and added checks for the dimensions on the c++ side but this seems for some reason not be active. 

This pull request is a test that fails when `TightBinding` accepts incorrect hopping dimensions without throwing.

Please merge.

### DIFF
--- a/test/pytriqs/base/CMakeLists.txt
+++ b/test/pytriqs/base/CMakeLists.txt
@@ -35,6 +35,7 @@ add_python_test(gf_transpose)
 
 add_python_test(issue460)
 add_python_test(issue437)
+add_python_test(bug_tight_binding)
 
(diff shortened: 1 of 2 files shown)


parcollet:
The test should pass, it works on my OS X.
Tentatively rebasing and merging on unstable, we will see what the CI says...

PR #551, opened 2018-03-10
[test] gf handling of scalar and matrices test

HugoStrand:
There is an inconsistency in the API how Green's functions are modified when scalars on the RHS of the `<<` operator. Here is a small test case that breaks on this inconsistent treatment.

In short for a matrix valued Green's function the operations
```python
A << scalar + matrix
```
and
```python
A << scalar
A << A + matrix
```
gives different results.

This pull request contains a test that breaks when these two operations are not the same.

Please merge.

The reason the two operations are different is due to that the Gf implicitly interprets a scalar as a scalar times the unit matrix. This does not work when using `numpy` in the right hand side as in the first case, where numpy`s broadcasting rules are used to evaluate the "scalar + matrix" expression element wise.

I would suggest that the special treatment of a scalar should be removed.

Best, Hugo

### DIFF
--- a/test/pytriqs/base/CMakeLists.txt
+++ b/test/pytriqs/base/CMakeLists.txt
@@ -36,6 +36,8 @@ add_python_test(gf_transpose)
 add_python_test(issue460)
 add_python_test(issue437)
 
+add_python_test(bug_gf_scalar_matrix)
+
 # Add evaluator for g
 add_python_test(gf_eval)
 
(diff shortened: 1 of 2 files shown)


HugoStrand:
Dear all, this pull request is pending since four months. A short comment would be nice. Best, Hugo

parcollet:
There is no way to fix this. matrix + scalar in numpy  does element-wise addition
while TRIQS matrix use the mathematics convention (scalar = unit matrix).

PR #555, opened 2018-03-29
[lattice] Wanner90 reader for tight binding

HugoStrand:
This is a small addition to the tight binding class that enables one to read real-space hoppings from Wannier90 and build a Triqs/TBLattice object from that.

Note that this is very different from the DFTTools Wannier90 parser that only reads an already k-space discretized H(k) hopping matrix from file. The additional feature from reading the real-space hopping is the ability to freely choose the k-point discretization at a later point.

Please consider merging.


mzingl:
Hi,
 
actually the DFTTools Wannier90 parser reads in the H(r) and does a Fourier transform. (Sorry Hugo, for not pointing this out earlier...) However, this is clearly something which should be in the TB module. We can see how this can be directly used for the Wannier90 converter in DFTTools.

Best,
Manuel

mzingl:
Hi,

is there any chance of adding this feature to TRIQS? I was using your pull request heavily over the last month and I think it is a very useful addition to lattice tools. Further, I think that then the DFTTools wannier90 converter could be rewritten such that it builds on lattice tools.

Best,
Manuel

PR #558, opened 2018-04-05
[clef] bug in imtime bubble

HugoStrand:
Dear All,

This is a failing test for the bubble diagram computed in imaginary time `chi(tau) = g(tau) * g(-tau)`. When using `clef` to compute this product the end points of the imaginary time interval are wrong, since the evaluator for `gf<imtime>` defines evaluation at `tau=beta` as evaluation at `tau=0^-`.

After the failing test there is an example `hack` implementation that shifts the time when it is on the boundary and gives the expected correct result.

To fix this we would have to modify how the `imtime` evaluator works on the boundaries.

Is anyone depending on the current behavior at the boundaries?

Please merge this test so that we will remember fixing this in one way or another.

Best, Hugo

### DIFF
--- a/test/triqs/gfs/gf_bubble.cpp
+++ b/test/triqs/gfs/gf_bubble.cpp
@@ -1,0 +1,61 @@
+
+// Bug # : Clef expression for the bubble diagram in imaginary time
+
+#include <triqs/test_tools/gfs.hpp>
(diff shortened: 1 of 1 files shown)


parcollet:
Was fixed and merged in 2.0.

PR #568, opened 2018-05-11
[many_body_operator] Fix #567

### LINKED ISSUE #567: Many_body operators with generic fields
I am trying to use the c++ ```many_body_operator_generic``` class with a user-defined ```ScalarType```. Specifically, I am using the GiNaC library to have a symbolic ```ScalarType```, but the issue I found is more general. It appears that in some parts the code assumes that the identity elements of ```ScalarType``` is (possibly automatically) converted from the numerical value 1.0. For example, around line 155 of many_body_operator.hpp, method ```make_canonical``` :

```
   auto m = monomial_t{canonical_ops_t{is_dag, indices}};
   res.monomials.insert({m, scalar_t(1.0)});
   return res;
```

If my ```ScalarType``` is a class for rational number, the above code may not work at all (because it cannot convert a floating number to an exact fraction), or it may "degrade" the coefficients to a floating point number (if ```ScalarType``` is allowed to be either an exact rational or a floating point number).

Similar issues are found in imperative_operator.hpp, around line 223

```
#ifdef GCC_BUG_41933_WORKAROUND
    target_st(ind) += amplitude * apply_if_possible(M.coeff,args_tuple) * (sign_is_minus ? -1.0 : 1.0);
#else
    target_st(ind) += amplitude * apply_if_possible(M.coeff,args...) * (sign_is_minus ? -1.0 : 1.0);
#endif
```

and in space_partition.hpp on lines 74, 116

```
tmp_state(i) = amplitude_t(1.0);
```

and on line 225

```
initial_state(i) = amplitude_t(1.0);
```

I think a safer solution would be to replace everywhere the ```1.0``` with ```ScalarType(1)```, or, better, with the specific type-alias of the class, say, ```scalar_t(1)``` for the method ```make_canonical```, etc...

krivenko:
@francescopt Would this solve your problem?

### DIFF
--- a/triqs/hilbert_space/imperative_operator.hpp
+++ b/triqs/hilbert_space/imperative_operator.hpp
@@ -203,6 +203,8 @@ template <typename HilbertType, typename ScalarType = double, bool UseMap = fals
   auto args_tuple = std::make_tuple(args...);
 #endif
 
+  using amplitude_t = typename StateType::value_type;
+
   for (int i = 0; i < all_terms.size(); ++i) { // loop over monomials
    auto M = all_terms[i];
 #ifdef GCC_BUG_41933_WORKAROUND
@@ -219,9 +221,9 @@ template <typename HilbertType, typename ScalarType = double, bool UseMap = fals
     // update state vector in target Hilbert space
     auto ind = target_st.get_hilbert().get_state_index(f3);
 #ifdef GCC_BUG_41933_WORKAROUND
-    target_st(ind) += amplitude * apply_if_possible(M.coeff,args_tuple) * (sign_is_minus ? -1.0 : 1.0);
+    target_st(ind) += amplitude * apply_if_possible(M.coeff,args_tuple) * (sign_is_minus ? -amplitude_t(1) : amplitude_t(1));
 #else
-    target_st(ind) += amplitude * apply_if_possible(M.coeff,args...) * (sign_is_minus ? -1.0 : 1.0);
+    target_st(ind) += amplitude * apply_if_possible(M.coeff,args...) * (sign_is_minus ? -amplitude_t(1) : amplitude_t(1));
 #endif
    }); // foreach
   }
--- a/triqs/hilbert_space/space_partition.hpp
+++ b/triqs/hilbert_space/space_partition.hpp
@@ -71,7 +71,7 @@ template <typename StateType, typename OperatorType> class space_partition {
 
   // Iteration over all initial basis states
   for (index_t i = 0; i < size; ++i) {
-   tmp_state(i) = amplitude_t(1.0);
+   tmp_state(i) = amplitude_t(1);
    state_t final_state = H(tmp_state);
 
    // Iterate over non-zero final amplitudes
@@ -113,7 +113,7 @@ template <typename StateType, typename OperatorType> class space_partition {
 
   // Fill connection multimaps
   for (index_t i = 0; i < size; ++i) {
-   tmp_state(i) = amplitude_t(1.0);
+   tmp_state(i) = amplitude_t(1);
    auto i_subspace = subspaces.find_set(i);
 
    auto fill_conn = [this,i,i_subspace,store_matrix_elements]
@@ -222,7 +222,7 @@ template <typename StateType, typename OperatorType> class space_partition {
   // Iteration over all initial basis states
   for (index_t i = 0; i < tmp_state.size(); ++i) {
    state_t initial_state = tmp_state;
-   initial_state(i) = amplitude_t(1.0);
+   initial_state(i) = amplitude_t(1);
    auto i_subspace = subspaces.find_set(i);
 
    state_t final_state = op(initial_state);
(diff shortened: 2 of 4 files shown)


francescopt:
I copied the header files ```hilbert_space/imperative_operator.hpp```, ```hilbert_space/space_partition.hpp```, ```hilbert_space/state.hpp```, ```operators/many_body_operator.hpp``` from the unstable branch and patched them. It works now fine, with no rounding, thank you!

I also noticed a small change in the output. In the unstable branch printing an operator does not include a product symbol "*" between the coefficient and the operator part of a monomial. A code like

```
cout << 0.5 * c_dag(1) << endl;
```

gives output

```
0.5C^+(1)
```

while in the stable branch the output would be

```
0.5*C^+(1)
```

I think it would be better to include the "*" in the output, especially if ```ScalarType```is a symbol.

krivenko:
> I also noticed a small change in the output. In the unstable branch printing an operator does not include a product symbol "*" between the coefficient and the operator part of a monomial.

This looks like a problem with your installation. Try to wipe the installation directory, build the `unstable` branch and install it anew.

francescopt:
I am having some troubles in installing the unstable branch.
Cmake complained that it could not find cpp2py. So, first I have installed the cpp2py package on a subdirectory on my home. I needed to manually indicate to cmake the position of the installation of cpp2py with the option ```-DCpp2Py_DIR=directory_installation_cpp2py/share/cmake```. Then, the configuration went fine, but the compiler gives error, reporting that it cannot find ```cpp2py.hpp```. So I cleared everything and ran again cmake with

```
LDFLAGS="-L directory_installation_cpp2py/lib" CXXFLAGS="-I directory_installation_cpp2py/include" CFLAGS="-I directory_installation_cpp2py/include" cmake .....
```

Then, I get another error: in file ```.../pytriqs/lattice/lattice_tools_desc.py``` "no module named cpp2py.wrap_generator".
I am not very proficient in python, can you tell me if I should add some python-equivalent of CFLAGS to the cmake command line?
Better: shouldn't the cmake configuration set all the required paths after you give the option Cpp2Py_DIR?

krivenko:
>  Better: shouldn't the cmake configuration set all the required paths after you give the option Cpp2Py_DIR?

The installation procedure has changed in `unstable`. You should import cpp2py environment variables by running `source $INSTALL_PREFIX/share/cpp2pyvars.sh`.

Installation in full detail: https://triqs.github.io/triqs/unstable/install.html

@Wentzell `README.txt` on the `unstable` branch still refers to the old documentation web-site.

francescopt:
Thanks for the help. I followed the instruction, but still the unstable branch does not compile. Cmake configuration is fine, except for a warning:

```
CMake Warning:
  Manually-specified variables were not used by the project:

    PYTHON_INTERPRETER
```

In fact, for the stable branch, I had to manually set ```PYTHON_INTERPRETER``` to point to python2.

On ```make``` I get a compilation error. The first error is

```
[ 11%] Building CXX object pytriqs/gf/CMakeFiles/meshes.dir/meshes_wrap.cxx.o
/........./triqs.build_unstable/pytriqs/gf/meshes_wrap.cxx:863:21: error: template argument 1 is invalid
     gf_mesh<legendre> * _c;
```

I am using gcc 8.1. The stable branch compiles with no problem, except for the issue #569 I opened.

Wentzell:
> [ 11%] Building CXX object pytriqs/gf/CMakeFiles/meshes.dir/meshes_wrap.cxx.o
> /........./triqs.build_unstable/pytriqs/gf/meshes_wrap.cxx:863:21: error: template argument 1 is invalid
>      gf_mesh<legendre> * _c;

Thats strange. I have tried with gcc 8.1 but I ran only into different problems later in the build procedure. 

Either way the fact is gcc 8.1 does currently not compile triqs/unstable!

The problem is that they changed the way that the 'math.h' header operates (compare [old](https://gcc.gnu.org/onlinedocs/gcc-7.3.0/libstdc++/api/a00011_source.html) vs [new](https://gcc.gnu.org/onlinedocs/gcc-8.1.0/libstdc++/api/a00011_source.html)).

By changing from

```c++
#if __STDCPP_WANT_MATH_SPEC_FUNCS__ == 1
```

to 
```c++
#if _GLIBCXX_USE_STD_SPEC_FUNCS
```

all the math special functions, including ```std::beta```, ```std::legendre``` and others, are moved up into the global namespace when you compile in c++17 mode. This is problematic for any physics c++ code I guess.

This means we have to completely get rid of any `math.h` header inclusions (`<cmath>` seems to be _mostly_ fine in that regard). The current origins are inclusion via `clef/adapters/math.hpp` `mc_tools/mc_generic.hpp` and `mc_tools/random_generator.hpp`, which is easily fixed, but also via  `Python.h`, which is unfortunately a bigger problem ...

francescopt:
> If you don't really plan to use unstable, I can open another pull request that fixes #567 on master.

Thank you, that would be better; I prefer to stick to the master branch.

> I would suggest opening a separate issue.

I agree. Should I open a new issue myself?

PR #572, opened 2018-05-18
[many_body_operator] Fix #567 (master)

### LINKED ISSUE #567: Many_body operators with generic fields

krivenko:
@francescopt Here the fix for #567 to be merged into `master`.

### DIFF
--- a/triqs/hilbert_space/imperative_operator.hpp
+++ b/triqs/hilbert_space/imperative_operator.hpp
@@ -203,6 +203,8 @@ template <typename HilbertType, typename ScalarType = double, bool UseMap = fals
   auto args_tuple = std::make_tuple(args...);
 #endif
 
+  using amplitude_t = typename StateType::value_type; 
+
   for (int i = 0; i < all_terms.size(); ++i) { // loop over monomials
    auto M = all_terms[i];
 #ifdef GCC_BUG_41933_WORKAROUND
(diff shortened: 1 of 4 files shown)


francescopt:
Thank you, this solves the issue #567 .

parcollet:
Fixed and merged in 2.0

PR #574, opened 2018-05-23
[test] gf_density boson test fails! n = infty...

HugoStrand:
The density calculation is completely broken for bosons and there is no test that checks this functionality.. here is a test.

Please, merge.


HugoStrand:
I have implemented the density calculation for the bosonic case in the last commit. Please, review and merge.

PR #585, opened 2018-06-01
Add key-word argument for center of SemiCircular

DerWeh:
Add `chem_potential` key-word argument to allow SemiCircular density of states away from half-filling.
The argument should be renamed to match TRIQS naming conventions.

The function seems not yet to properly work, as `density()` complains about non-vanishing 0th.

### DIFF
--- a/pytriqs/gf/descriptors.py
+++ b/pytriqs/gf/descriptors.py
@@ -71,21 +71,29 @@ class SemiCircular (Base):
       
     (Only works in combination with frequency Green's functions.)
     """
-    def __init__ (self, half_bandwidth):
+    def __init__ (self, half_bandwidth, chem_potential=0.):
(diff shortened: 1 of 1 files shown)


DerWeh:
Personally I would prefer this version of the function:
```python
om_rel = om / D
return 2./D*om_rel*(1. - (1. - 1./(om_rel*om_rel))**0.5)
```

While it requires a complex root and and a devision, it is far more readable to me, as the units are clear. I don't know if that is an issue concerning numerical accuracy.
`**0.5` avoids the difficulty if `float sqrt`, `complex sqrt` or `numpy sqrt` is needed.

parcollet:
Merging this.
For the function writing, my choice was to avoid complex sqrt since I was not sure it would 
be clearer to think about the precise determination of sqrt in the complex plane.

PR #587, opened 2018-06-05
Move methaclass of `pytriqs.gf.Gf` after docstring

DerWeh:
Docstrings have to be the first statement after the definition.


DerWeh:
You might want to consider [Pydocstyle](http://www.pydocstyle.org/en/latest/) for consitent doc strings.

PR #589, opened 2018-06-06
Dev bz mesh h5 bug

HugoStrand:
The `BrillouinZoneMesh` does not write and read its domain `BrillouinZone` to hdf5 so whenever it is read from disk the reciprocal basis vectors are initialized to unity.

This patch adds read/write steps in the h5_read/h5_write methods for the domain.

Please merge.


Wentzell:
Working on it!

Wentzell:
Merged in 741489ba349099386dbe80681c51e494e001217f

PR #610, opened 2018-07-18
[test] analytic tail coeff

HugoStrand:
Dear all,

Here is a small test with a sanity check for the tail fit. It compares the tail fit high frequency coefficients with the exact coefficients on a simple Green's function 

G(iw) = 1/(iw - 1) 

with known analytic high frequency moments. Its high frequency expansion is simply G(iw) = \sum_k 1/(iw)^k.

I think that fitting the tail of this function should be a feature of the tail fit.

Please merge.


parcollet:
@Wentzell 

Nils, is this still relevant ? 
If not, can we remove it ?

Wentzell:
This is incorporated in the tail_issues test added in 5bcaf4b8a8e98bd1c3d617145e7a326c06fa410f

Issue #621, opened 2018-08-29
Iteration over empty array segfaults


HugoStrand:
Dear all,

I am writing some generic code with mpi parallelization and some ranks happen to have zero length arrays to work with. However, this is not very pleasant since looping over an empty `triqs::array` segfaults.

Here is a minimal example:
```c++
  int size = 0;
  array<double, 1> arr(size);
  for( auto i : arr ) std::cout << i << "\n";
```

Is this a bug?

Best, Hugo

running this gives
```
[ RUN      ] Array.empty_iteration
[void:03830] *** Process received signal ***
[void:03830] Signal: Segmentation fault: 11 (11)
[void:03830] Signal code: Address not mapped (1)
[void:03830] Failing at address: 0x0
[void:03830] [ 0] 0   libsystem_platform.dylib            0x00007fff60db3f5a _sigtramp + 26
[void:03830] [ 1] 0   ???                                 0x00000001187e92d2 0x0 + 4705915602
[void:03830] [ 2] 0   libgtest.dylib                      0x000000010d9bd378 _ZN7testing8internal35HandleExceptionsInMethodIfSupportedINS_4TestEvEET0_PT_MS4_FS3_vEPKc + 72
[void:03830] [ 3] 0   libgtest.dylib                      0x000000010d9bd23d _ZN7testing4Test3RunEv + 1197
[void:03830] [ 4] 0   libgtest.dylib                      0x000000010d9bee64 _ZN7testing8TestInfo3RunEv + 1188
[void:03830] [ 5] 0   libgtest.dylib                      0x000000010d9bf787 _ZN7testing8TestCase3RunEv + 311
[void:03830] [ 6] 0   libgtest.dylib                      0x000000010d9c8e5a _ZN7testing8internal12UnitTestImpl11RunAllTestsEv + 2090
[void:03830] [ 7] 0   libgtest.dylib                      0x000000010d9c8488 _ZN7testing8internal35HandleExceptionsInMethodIfSupportedINS0_12UnitTestImplEbEET0_PT_MS4_FS3_vEPKc + 72
[void:03830] [ 8] 0   libgtest.dylib                      0x000000010d9c83f0 _ZN7testing8UnitTest3RunEv + 176
[void:03830] [ 9] 0   empty_array_loop                    0x000000010d80a25b main + 75
[void:03830] [10] 0   libdyld.dylib                       0x00007fff60aa5015 start + 1
[void:03830] *** End of error message ***
Segmentation fault: 11
```

[...]

PR #622, opened 2018-08-29
[array] zero length iterator bugfix

HugoStrand:
Here is a bug fix for issue #621. Please consider merging. Best, Hugo


parcollet:
Code ok. Will merge after minor cleaning of test.

PR #623, opened 2018-08-30
[util] time_pt h5 serialization

HugoStrand:
Adding serialization for time_pt. (Required to be able to broad cast the cthyb solver over mpi in python.)

Please merge

Best, Hugo


parcollet:
string const& as argument ...

HugoStrand:
Fixed. Could you please have a second look? Best, Hugo

HugoStrand:
+1 month and 9 days... adding @Wentzell on the request list.

PR #631, opened 2018-09-05
[det_manip] expose precision warning and error bounds

HugoStrand:
This exposes the error and warning bounds used by the `det_manip` without breaking the api. This is used in cthyb and exposed as parameters (for power users).

Please review and merge.
Best, Hugo


parcollet:
change declaration to void check_mat_inv()
it is private, and used only once.

HugoStrand:
`check_mat_inv()` is now fixed. Thank you for the feedback. Can we merge?

PR #634, opened 2018-09-14
Total occupation number constraints in atom diag (fix of functionality regression)

HugoStrand:
Dear all,

This fixes the `atom_diag` feature of only considering a restricted range of the total occupation when building the Fock space.

The commit 8b28b5a exposes the interface and adds a test that uses the constrained occupation range constructor. The commit 64ffc6 is a patch from @krivenko that fixes the internal bug in `atom_diag` that caused it to segfault when encountering empty subsectors. With this modification the test passes.

@Wentzell I discussed the test and interface with you before. Here is both that and the required fix to make the test pass. Could you please have a look?

Best regards,
Hugo

### DIFF
--- a/test/triqs/atom_diag/f_shell.cpp
+++ b/test/triqs/atom_diag/f_shell.cpp
@@ -70,16 +70,13 @@ TEST(atom_diag_real, f_shell_kanamori) {
   
   auto H = make_hamiltonian<many_body_operator_real>(n_orb, mu, U, J);
 
-  //int n_min = 0, n_max = 2 * n_orb; // works
-  int n_min = 0, n_max = 3*2; // breaks
+  int n_min = 0, n_max = 3*2;
 
   auto ad = triqs::atom_diag::atom_diag<false>(H, fops, n_min, n_max);
 
   std::cout << "Found " << ad.n_subspaces() << " subspaces." << std::endl;
-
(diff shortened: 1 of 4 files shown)


test/triqs/atom_diag/f_shell.cpp
```diff
@@ -0,0 +1,85 @@
...
+  double mu = 0.5*U;
+  
+  auto H = make_hamiltonian<many_body_operator_real>(n_orb, mu, U, J);
+
+  //int n_min = 0, n_max = 2 * n_orb; // works
+  int n_min = 0, n_max = 3*2; // breaks
+
+  auto ad = triqs::atom_diag::atom_diag<false>(H, fops, n_min, n_max);
+
+  std::cout << "Found " << ad.n_subspaces() << " subspaces." << std::endl;
```
parcollet: Transform into an EXPECT

test/triqs/atom_diag/f_shell.cpp
```diff
@@ -0,0 +1,85 @@
...
+
+  int op_linear_index = 0;
+  int sidx = 1000; // subspace index
+  std::cout << ad.c_matrix(op_linear_index, sidx) << "\n";
```
parcollet: Add expect

Issue #626, opened 2018-08-31
Thread-safe gf_views wanted


HugoStrand:
Dear all,

I am trying to work with `triqs` greens functions and `openmp` threading and have issues with the `gf_view` not being thread safe. The `array` seems not to suffer from these issues. Here is a minimal example that give me intermittent segfaults due to some race condition inside the views:
```c++
TEST(Array, openmp_gf_imfreq_imfreq) {
  auto _ = all_t{};
  double beta = 1.;
  int nw = 500;
  int nb = 2;
  gf_mesh<imfreq> m{beta, Fermion, nw};
  gf<cartesian_product<imfreq, imfreq>> g1({m, m}, {nb, nb});
  gf<cartesian_product<imfreq, imfreq>> g2({m, m}, {nb, nb});
  g1 *= 0.;
  g2 *= 0.;
  
#pragma omp parallel for 
  for (int idx = 0; idx < m.size(); idx++) {
    auto iter = m.begin(); iter += idx; auto w = *iter;
    g1[w, _] = g2[w, _];
  }
  EXPECT_GF_NEAR(g1, g2);
}
```
It would be extremely useful if there was an option to get "non-owning" views on triqs Green's functions. Right now I have to add `#pragma omp critical` sections every time I store or read the values from a larger Green's function.

Please help

Best regards,
Hugo

here is one segfault output:
```
[ RUN      ] Array.openmp_gf_imfreq_imfreq
[ccqlin009:329644] *** Process received signal ***
[ccqlin009:329644] Signal: Segmentation fault (11)
[ccqlin009:329644] Signal code: Address not mapped (1)
[ccqlin009:329644] Failing at address: 0x7fd821330010
[ccqlin009:329644] [ 0] /lib64/libpthread.so.0(+0xf5e0)[0x7fd8b3b595e0]
[ccqlin009:329644] [ 1] ./test/triqs/arrays/openmp(_ZN5triqs3gfs20assert_gfs_are_closeINS0_2gfINS0_17cartesian_productIJNS0_6imfreqES4_EEENS0_13matrix_valuedEEES7_EEvRKT_RKT0_d+0x231)[0x41d331]
[...]

PR #644, opened 2018-10-18
[array] make ref_counters std::atomic

HugoStrand:
Fix to make gf_views thread safe using std::atomic on the reference counters.

### DIFF
--- a/triqs/arrays/storages/mem_block.hpp
+++ b/triqs/arrays/storages/mem_block.hpp
@@ -3,6 +3,7 @@
  * TRIQS: a Toolbox for Research in Interacting Quantum Systems
  *
  * Copyright (C) 2011 by O. Parcollet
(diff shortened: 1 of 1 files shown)


parcollet:
That is what I meant indeed. Does it fix the issue ?

parcollet:
PS : 
the concern here can be performance. Years ago, I ruled out using std::shared_ptr, 
which does exactly that, because it degraded performance in some cases.
But I was using shared_view way too much, so maybe it is not relevant any more...
To be checked.

HugoStrand:
This fixes the segfaults in #626 . I have no idea about performance, but taking views should imho be thread safe.

HugoStrand:
The errors from the IC are now fixed. Thanks @dylex ! @parcollet can we merge?

PR #647, opened 2018-11-14
[lapack] replaces fortran wrappers

phdum:
- pull in headers from blas/lapack reference implementation (https://github.com/Reference-LAPACK)
- hides fortran includes of lapack functions in cpp binary file
- cleans up header files for function on triqs arrays
- move functions from blas to lapack namespace

### DIFF
--- a/test/triqs/arrays/gtsv.cpp
+++ b/test/triqs/arrays/gtsv.cpp
@@ -48,21 +48,21 @@ TEST(gtsv, dgtsv) {
     auto dl(DL);
     auto d(D);
     auto du(DU);
-    blas::gtsv(dl, d, du, B1);
+    lapack::gtsv(dl, d, du, B1);
     EXPECT_ARRAY_NEAR(B1, ref_sol_1);
   }
   {
     auto dl(DL);
     auto d(D);
     auto du(DU);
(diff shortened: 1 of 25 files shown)


parcollet:
1)
Looks very good.
The only unfortunate part is that we now have two calls, 
f77::gemm, then f77_gemm, that the compiler can not inline ...
For large matrix, it should be irrelevant.
For small matrix, blas/lapack may not be optimal anyway.

2)
Small details : 

a)pass clang-format before commit.
clang-format -i **/*.?pp

@Wentzell : could you give both of us the git hook to have this done at commit time, please ?

b) in gtsv .hpp , a strange comment : 
+// TODO: MOVE THIS TO ::LAPACK
+namespace triqs::arrays::lapack {

parcollet:
Approved after minor changes are done (formatting).

Wentzell:
Merged in a4d9982f and corrected in ce723748

For the commit hooks you can use the file [here](https://github.com/andrewseidl/githook-clang-format/blob/master/clang-format.hook), but I find it much cleaner to set it up in the editor with a clang-format on save. This way you can still separate your formatting changes from your relevant changes.

Issue #604, opened 2018-07-10
Monte Carlo code spending half of its time calling the clock


amoutenet:
Hi all,

I am currently writing a Monte Carlo code having very quick updates, and when profiling it I noticed that the code is spending 60% (!) of the time calling the now() function of the clock. I am using gcc7 and working on the baf01f commit of the unstable branch.

Looking more carefully at this problem, we noticed with Michel that this is coming from the three occurences of now() in the triqs/utility/timer.hpp header that I copy-paste below for clarity (lines of the occurences are followed by //comment, see below):
```c++
#pragma once

#include <chrono>

namespace triqs {
  namespace utility {

    class timer {
      public:
      using clock_t = std::chrono::high_resolution_clock;

      private:
      clock_t::time_point start_time;
      clock_t::duration total_time = clock_t::duration(0);
      bool running                 = false;

      public:
      void start() {
        running    = true;
        start_time = clock_t::now(); //comment
      }
      void stop() {
        total_time += clock_t::now() - start_time; //comment
        running = false;
      }
      bool is_running() const { return running; }
      operator double() const {
        std::chrono::duration<double> total_time_seconds(total_time);
        if (is_running()) total_time_seconds += clock_t::now() - start_time; //comment
        return total_time_seconds.count();
      }
    };
  } // namespace utility
} // namespace triqs
```
I tried to replace std::chrono::high_resolution_clock by std::chrono::system_clock and std::chrono::steady_clock but it didn't change anything. The only workaround we found is to comment the three lines involving now(). Fortunately, the callbacks utility is not affected by this procedure and I can still give a precise run time to my Monte Carlo.

I understand this is very deep in the standard c++ library and there is no easy solution to this problem, but I think it is good to be aware of it. 
What I would suggest is the following:
- put timing of measures optional (in my case for instance it is just the MC weight, so virtually 0)
- put the percentage optional, maybe just keeping three of them, one every 25% for instance (because it is very useful to roughly now how long a calculation will take, especially when tuning parameters before going into production mode)
- reduce as much as possible the other occurences of these calls

Cheers,
Alice

parcollet:
Thank you Alice for analysing and reporting this.

Indeed, the clock is a system call, it will be slow whatever you do.
The timer call should not be in the inner QMC loop.

@HugoStrand @Wentzell : can one of you take care of this ?

HugoStrand:
Dear Alice,

[...]

PR #650, opened 2018-11-20
[mc] make measure timer optional

HugoStrand:
This is a fix for issue #604 making it possible to disable the timers for the mc measures by passing an additional flag in the constructor. As suggested in the triqs meeting discussion Nov 16.

@amoutenet Could you please try this fix and check that it solves your speed issue?

Best, H

### DIFF
--- a/triqs/mc_tools/mc_generic.hpp
+++ b/triqs/mc_tools/mc_generic.hpp
@@ -119,9 +119,9 @@ namespace triqs {
    * @param name                     Name of the measure
    *
    */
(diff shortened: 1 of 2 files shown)


parcollet:
@amoutenet @mferrero 
Hi,
The change look good to me.
Can one of you quickly check that the issue is gone now ?

Thanks, 

Olivier

amoutenet:
@HugoStrand @parcollet I tested the changes, is it now good to merge?
Alice

PR #653, opened 2018-11-28
[detmanip] Don't regenerate on singular det, Throw in regeneration

Wentzell:
-det_manip should never explicitly go through a singular matrix
-singular matrices should be SKIPPED OVER by the proper operations,
e.g. try_insert2 instead of 2x (try_insert + complete operation)
-throw exception if we encounter a singular determinant in the
regeneration
-regenerate if determinant is zero in the accessor function
-minor formatting


Wentzell:
> Why did we not merge this ?
It was just awaiting review.

> Shall we put it for 2.3 in unstable ?
Yes, that seems like the best course of action!

Issue #675, opened 2019-02-01
Bug: Operator(0.0) not hdf5 serializable


HugoStrand:
Dear all,

I just hit a bug in the serialization of the operator class. Trying to store a zero operator segfaults. Here is a minimal example:
```python
from pytriqs.operators import Operator
from pytriqs.archive import HDFArchive

H_int = Operator(0.0)

with HDFArchive('H_int.h5', 'w') as arch:
    arch['H_int'] = H_int
```
This crash and burns with:
```
Traceback (most recent call last):
  File "bug_operator.py", line 8, in <module>
    arch['H_int'] = H_int
  File "/Users/hugstr/apps/triqs_unstable/lib/python2.7/site-packages/pytriqs/archive/hdf_archive.py", line 184, in __setitem__
    val.__write_hdf5__(self._group,key)
RuntimeError: .. Error occurred at Fri Feb  1 10:20:02 2019

.. Error in h5 writing of object Operator
.. C++ error was : 
.. Triqs runtime error at /Users/hugstr/dev/triqs/triqs/h5/base.cpp : 49

Cannot set hyperslab
.. Error occurred on node .. C++ trace is : 
    frame #6: 0x00000001056348cc libtriqs.dylib`triqs::h5::dataspace_from_LS(int, bool, unsigned long long const*, unsigned long long const*, unsigned long long const*, unsigned long long const*) + 652
    frame #7: 0x000000010564325d libtriqs.dylib`triqs::h5::h5_write_attribute(long long, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&, std::__1::vector<std::__1::vector<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >, std::__1::allocator<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > > >, std::__1::allocator<std::__1::vector<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >, std::__1::allocator<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > > > > > const&) + 797
    frame #8: 0x0000000105649cd8 libtriqs.dylib`triqs::hilbert_space::h5_write_attribute(long long, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&, triqs::hilbert_space::fundamental_operator_set const&) + 776
[...]

PR #677, opened 2019-02-02
[h5] Fix issue #675

krivenko:
This commit makes

`void h5_write_attribute(hid_t ob, std::string const & name, std::vector<std::vector<std::string>> const & V)`
`void h5_read_attribute(hid_t ob, std::string const & name, std::vector<std::vector<std::string>> & V)`

properly handle corner cases when some of the vectors/strings are empty. `h5_io` test has been extended to cover the changed functions.

### DIFF
--- a/test/triqs/h5/h5_io.cpp
+++ b/test/triqs/h5/h5_io.cpp
@@ -141,4 +141,44 @@ TEST(H5Io, Vector) {
   EXPECT_EQ(mv, mmv);
 }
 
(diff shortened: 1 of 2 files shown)


Wentzell:
I did some further simplifications. @krivenko Can you review and comment/merge?

krivenko:
@Wentzell LGTM. However, I don't think I have permissions to either approve or merge.

PR #681, opened 2019-02-20
Removed deprecation warning from the latest version of h5py

shinaoka:
As the title says, the use of deprecated "value" and "h5py.highlevel" has been removed.


Wentzell:
Hi @shinaoka,

Thanks for pointing this out!
This was already mostly addressed in fc9449e8 and b99a0674
I have moved these changes also to the 1.4.x and 2.0.x branches together with the ```A.value``` change.

Issue #342, opened 2016-06-29
block_gf: design problem?


tayral:
Hi,
Consider the following code snippet:

``` c++
  auto Ga = gf<imfreq>({1, Fermion}, {1, 1});
  auto Gb = gf<imfreq>({1, Fermion}, {1, 1});
  triqs::clef::placeholder<0> w_; 
  Ga(w_) << 1 / (w_ + 2); 
  Gb(w_) << 1 / (w_ - 2); 

  auto B_ab = make_block_gf<imfreq>({"a", "b"}, {Ga, Gb});
  auto B_ba = make_block_gf<imfreq>({"b", "a"}, {Gb, Ga});

  auto B_res = B_ab;
  for(auto const & ind : B_res.mesh())    B_res[ind] = B_ba(ind);
```

What should one expect this code to do?
It seems to me that the last line should mean that there is a block-by-block equality between B_res and B_ba, namely (in a python syntax) B_res["a"] = B_ba["a"] = Ga and similarly for "b".
Yet, this is not what happens, because what the last line really does is interpret "ind" as an integer indexing the blocks, so that one ends up with B_res["a"] = B_ba["b"] = Gb... which I find quite surprising... and error-prone.
What do you think?

parcollet:
@mferrero : requires a discussion

krivenko:
Each of these two lines
```c++
auto B_ab = make_block_gf<imfreq>({"a", "b"}, {Ga, Gb});
auto B_ba = make_block_gf<imfreq>({"b", "a"}, {Gb, Ga});
```
assigns different integer indices to `"a"` and `"b"`. What if we make the constructor assign those indices according to the lexicographical order of strings in `{"b", "a"}`?
`B_ab` and `B_ba` will be completely equivalent objects, just as they should.

tayral:
Note: also raised in #349. A quite important issue!

krivenko:
This annoying issue still exists in `unstable`...

krivenko:
Let me ask again: why not sort block names in the constructor?
It is frustrating that `assert_block_gfs_are_close()` remains unusable due to this issue...

parcollet:
Indeed, we were precisely discussing this with @Wentzell 
@mferrero : opinion ?

Wentzell:
Indeed!
When creating the block gf we should sort lexicographically w.r.t the block_name
(and then the block2_name for block2_gf).

mferrero:
[...]

PR #686, opened 2019-03-06
Adjust fundamental operator set to respect order in gf_struct #342

Wentzell:
In accordance with the type-change of ```gf_struct_t``` to
```c++
gf_struct_t = vector<pair<int,vector<indices_t>>>
```
that was introduced in a746ba5 this commit adjusts the 
```fundamental_operator_set``` to respect the order of blocks in gf_struct

This PR still contains ***unfixed test failures*** for the atom_diag part of triqs.

@krivenko Could you comment on/review this PR and have a look at the test failures in atom_diag?

### DIFF
--- a/test/triqs/atom_diag/hamiltonian.hpp
+++ b/test/triqs/atom_diag/hamiltonian.hpp
@@ -9,10 +9,8 @@ using namespace triqs::operators;
 // Prepare funcdamental operator set
 fundamental_operator_set make_fops() {
   fundamental_operator_set fops;
-  for (int o : range(3)) {
-    fops.insert("up", o);
-    fops.insert("dn", o);
-  }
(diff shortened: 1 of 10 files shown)


krivenko:
I'll look at the code a bit later...

Now I only have a quick question about `gf_struct_t`.
The commit message of a746ba5 states that `gf_struct_t = vector<pair<string,vector<indices_t>>>`. Who is to believe?

Wentzell:
You are indeed correct. The type is  ```vector<pair<string,vector<indices_t>>>```

Wentzell:
Hey @krivenko 
Thank you for the test fixes and comments.
I have addressed them in 985898dc

Let me know if we can merge it like this.

krivenko:
LGTM, thank you!

Issue #560, opened 2018-04-09
How to do a OpenMP parallel for loop over gf_mesh<imfreq>?


HugoStrand:
Dear all,

I am trying to make a little OpenMP parallelization of a loop over a `gf_mesh<imfreq>` mesh. The original code uses the nice `c++17` syntax
```c++
for (auto const &k : kmesh) { ... }
```
Since OpenMP (4.0) does not yet support this range-based for loop syntax I have tried to use the iterator interface directly, i.e.
```c++
#pragma omp parallel for
for (auto k_iter = kmesh.begin(); k_iter != kmesh.end(); k_iter++ ) { 
  auto k = *k_iter;
  ... 
}
```
This does not compile since OpenMP requires the comparison operator to be `<` or `<=`. However that is not implemented by the iterator interface of `gf_mesh<imfreq>`. Trying with the `<` operator I get the error message:
```
error: invalid operands to binary expression
      ('triqs::gfs::mesh_pt_generator<triqs::gfs::cluster_mesh>' and 'cluster_mesh::const_iterator' (aka
      'mesh_pt_generator<triqs::gfs::cluster_mesh>'))
    for (auto q_iter = kmesh.begin(); q_iter < kmesh.end(); q_iter++ ) {
                                      ~~~~~~ ^ ~~~~~~~~~~~
/Users/hstrand/apps/triqs_unstable/include/triqs/./clef/adapters/../clef.hpp:205:2: note: candidate template ignored:
      requirement 'is_any_lazy<mesh_pt_generator<cluster_mesh> &, mesh_pt_generator<cluster_mesh> >::value' was not
      satisfied [with L = triqs::gfs::mesh_pt_generator<triqs::gfs::cluster_mesh> &, R =
      triqs::gfs::mesh_pt_generator<triqs::gfs::cluster_mesh>]
 TRIQS_CLEF_OPERATION(less,       <);
 ^
/Users/hstrand/apps/triqs_unstable/include/triqs/./clef/adapters/../clef.hpp:191:114: note: expanded from macro
      'TRIQS_CLEF_OPERATION'
  ...std14::enable_if_t<is_any_lazy<L, R>::value, expr<tags::TAG, expr_storage_t<L>, expr_storage_t<R>>> operator OP(L&& ...
                                                                                                         ^
/Users/hstrand/apps/triqs_unstable/include/triqs/arrays/./expression_template/../linalg/../vector.hpp:256:84: note: 
[...]

PR #691, opened 2019-03-29
Make itertools and range functionality standalone, Add mpi_slice range adapter

Wentzell:
This Pull Request achieves the following

- Major cleaning of itertools.hpp
- Move functionality for integer ranges from triqs::arrays namespace to triqs::utility and into separate range.hpp file
- Add a new range adapting function
```C++
mpi_scatter(Range R, triqs::mpi::communicator comm)
```
that scatters a range uniformly over all ranks of an mpi communicator.


Wentzell:
This PR fixes #560

Wentzell:
@parcollet Please review the updated PR. I have incorporated the comments as discussed

- Enforce inlining in itertools member functions
- Add ```slice(Range, start, end, step)``` range adaptor that will view only part of a range.
- Rename `mpi_scatter` -> `mpi_slice` and define it by means of `triqs::mpi::slice_range`
- I have also adjust `triqs::mpi::slice_range` and `triqs::mpi::slice_length` to match the API
 of `range(start, end, step)`, i.e. to exclude `end` from the range

Wentzell:
@parcollet I now moved the itertools / range functionality into its own itertools namespace as discussed
I further merged the range.hpp header into itertools.hpp header.
I made the itertools not depend on triqs and adjusted the cmake structure of the tests accordingly.
Along the way I did some cleaning of the CMakeFiles

PR #704, opened 2019-05-06
[bug] bcast mesh

HugoStrand:
This is a test for a bug in BZMesh where the dimension turns from 3 to 2 when using bcast on the mesh.

### DIFF
--- a/test/pytriqs/base/CMakeLists.txt
+++ b/test/pytriqs/base/CMakeLists.txt
@@ -42,6 +42,7 @@ add_python_test(issue460)
 add_python_test(issue437)
 add_python_test(issue649)
 add_python_test(bug_tight_binding)
+add_python_test(bug_bcast_mesh)
 
 # Issues with the high frequency moments
 add_python_test(tail_issues)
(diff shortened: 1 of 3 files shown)


HugoStrand:
This is a bug that have appeared after the 2.0.0 release. I have regularly been broadcasting input data read from disk by rank 0 on earlier versions of Triqs without problems.

The bug comes from the combination of the default constrictor of `bravais_lattice` that makes a two dimensional lattice, and the boost serialization implementation that omitted to propagate the dimension stored in `dim_`. This generated an inconsistent mesh with 3x3 units but with `dim_ = 2` after pickling in the `mpi.bcast` call.

The problem is fixed by adding `dim_` to the serialization. See the second commit.

Please merge.

HugoStrand:
Dear all,

This is an evident bug, with a one line fix, and a test. 
It has been waiting for 3 months, but should be a no-brainer.

Best, H

Wentzell:
@HugoStrand Thank you for the fix and test!

Wentzell:
Please create future PR's directly to the unstable branch. I will `git cherry-pick` this one over.

PR #718, opened 2019-06-21
Fix condition check in clang linker adjustment

phdum:
This fixes a tiny bug (probably copy-paste) in the linker adjustments for clang in the TRIQSConfig.cmake. This had the effect of adding the command "-LANDDEFINEDENV{LIBRARY_PATH}" when I linked TRIQS to a different program.

The same code is correct in the CMakeFile for TRIQS itself.


Wentzell:
This was indeed a copy paste error. Thanks for the fix!

PR #726, opened 2019-07-22
UBSAN sanitization build in jenkins and related fixes

dylex:
Check jenkins status before merging.

### DIFF
--- a/Dockerfile.build
+++ b/Dockerfile.build
@@ -11,7 +11,7 @@ ENV SRC=/src \
 
 COPY . ${SRC}/triqs
 WORKDIR ${BUILD}/triqs
-RUN chown build -R . ${SRC}/triqs
+RUN chmod -R +rX ${SRC}/triqs && chown build .
 USER build
 ARG BUILD_DOC=0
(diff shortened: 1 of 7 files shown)


dylex:
This is ready.

Wentzell:
Hey Dylan,

Thanks a lot!

Why do we build libc++ and libc++abi instead of using the ubuntu packages in that case?

https://packages.ubuntu.com/bionic/libc++-dev
https://packages.ubuntu.com/bionic/libc++abi-dev

dylex:
I just copied that from our existing ubuntu build, but probably a good point to eliminate duplication, at least.

I think that originally came from xenial which didn't have those packages at the time. We may be able to switch the ubuntu-clang build over now. I'll give it a try.

Wentzell:
@dylex Do you have an idea whats going wrong with the jenkins osx builds?

dylex:
Not exactly, but it looks unrelated to this stuff. It looks like it did some automatic updates and rebooted last night around that time, so maybe it was a shutdown process killing things. It's running fine now.

PR #727, opened 2019-07-26
Fix get_opt_mat function of atom_diag

amoutenet:
The new many-body op was already expressed in the eigenbasis, applying the unitary matrices again was expressing it in the Fock space.

### DIFF
--- a/triqs/atom_diag/atom_diag.hpp
+++ b/triqs/atom_diag/atom_diag.hpp
@@ -275,7 +275,7 @@ namespace triqs {
        *
        * Throws, in case the provided operator does not respect the block symmetries used in the diagonalization.
        */
-      op_block_mat_t get_op_mat(many_body_op_t const &op) const;
+      op_block_mat_t get_op_mat(many_body_op_t const &op, bool fock_space = true) const;
       
 
       private:
--- a/triqs/atom_diag/impl/atom_diag.cpp
+++ b/triqs/atom_diag/impl/atom_diag.cpp
@@ -112,7 +112,7 @@ namespace triqs {
 
     // -----------------------------------------------------------------
 
-    ATOM_DIAG_METHOD(op_block_mat_t, get_op_mat(many_body_op_t const &op) const) {
+    ATOM_DIAG_METHOD(op_block_mat_t, get_op_mat(many_body_op_t const &op, bool fock_space) const) {
       op_block_mat_t op_mat(n_subspaces());
 
       for(int b : range(n_subspaces()) ) {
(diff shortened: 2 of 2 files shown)


amoutenet:
As a side note, this is not specific to unstable and could be merged on 2.1.x (should I do another PR?)

HugoStrand:
Dear Alice,

This will most likely break TRIQS/cthyb sampling by insertion where the current functionality/basis choice is used, please see:
https://github.com/TRIQS/cthyb/blob/2.1.x/c%2B%2B/triqs_cthyb/impurity_trace.hpp#L206

Please fix that code or add a default valued flag that controls whether the transform is done or not and retains the previous behavior if not additional parameter is passed.

Cheers, Hugo

amoutenet:
Dear Hugo,
Indeed I missed that! I checked that the modification passed the cthyb tests but I guess this part might not be tested. I'll add a default value flag then.

Alice

HugoStrand:
Dear Alice, Ok! Thank you. Best, Hugo

amoutenet:
Added a fock_space = true defat boolean!

Wentzell:
@amoutenet Could you rebase your changes onto unstable so we can see if the CI passes there?

amoutenet:
Done! Let's see what comes out

amoutenet:
Just checked Jenkins output, it managed to build everything and passed all tests, @Wentzell do you understand why there is an error?

Wentzell:
@amoutenet Thank you for the rebase. The failure seems to be linked to a problem with the continuous integration.

@dylex Something seems to be going wrong with the h5py setup for the OSX Pull Request builds. Could you look into this?

https://jenkins.flatironinstitute.org/job/TRIQS/job/triqs/view/change-requests/job/PR-727/4/execution/node/80/log/

dylex:
This should be fixed. (OSX automatic updates keep removing xcode components.)

Wentzell:
@krivenko Could you have a quick look at this?

krivenko (approved):

Wentzell:
@HugoStrand Since the cthyb impurity trace works directly in the local eigenbasis you will also need the auxiliary operator matrices in that basis, or am I mistaken?
As Alice pointed out the previous version of the get_opt_mat was **not** returning the matrix in the eigen basis.
Did you ever do any tests beyond one-band models for the insertion sampling?

Wentzell:
> @HugoStrand Since the cthyb impurity trace works directly in the local eigenbasis you will also need the auxiliary operator matrices in that basis, or am I mistaken?
> As Alice pointed out the previous version of the get_opt_mat was **not** returning the matrix in the eigen basis.
> Did you ever do any tests beyond one-band models for the insertion sampling?

@HugoStrand  Can you comment on this?

Wentzell:
We have fixed this bug in f3ab183b3 and provided an easy to way to check if a calculation was previously effected.

By setting the environment variable `CHECK_ISSUE833` before running the code with e.g. `export CHECK_ISSUE833=1`
the user will get an exception in case the result was previously wrong.

PR #741, opened 2019-08-19
[cmake] Generate and install a triqs++ compiler wrapper

Wentzell:
Inspect cmake targets triqs and python_and_numpy recursively to extract full sets of compile and linktime flags. Generate and install a python-based compiler-wrapper 'triqs++'


Wentzell:
This was merged after review together with @parcollet

PR #744, opened 2019-08-27
Remove unnecessary modulo operations from index_to_linear of cluster_mesh

Wentzell:
* Explicitly check is_within_boundary for index_to_linear and index_to_point in Debug mode
* Print stack_trace in contract macros EXPECTS, ENSURES, ASSERT and disable without debugging
* Minor formatting corrections

### DIFF
--- a/test/triqs/gfs/gf_empty.cpp
+++ b/test/triqs/gfs/gf_empty.cpp
@@ -9,6 +9,6 @@ TEST(Gf, Exceptions) {
 
   auto g = gf<imtime, scalar_valued>{};
 
-  EXPECT_THROW(g[closest_mesh_pt(5.3)], triqs::arrays::key_error);
+  ASSERT_DEATH(g[closest_mesh_pt(5.3)], "");
 }
 MAKE_MAIN;
--- a/triqs/gfs/meshes/discrete.hpp
+++ b/triqs/gfs/meshes/discrete.hpp
@@ -51,9 +51,19 @@ namespace triqs {
       ///
       utility::mini_vector<size_t, 1> size_of_components() const { return {size()}; }
 
+      /// Is the point in the mesh ?
+      bool is_within_boundary(all_t) const { return true; }
+      bool is_within_boundary(index_t idx) const { return ((idx >= 0) && (idx < size())); }
+
       /// Conversions point <-> index <-> discrete_index
-      long index_to_point(index_t ind) const { return ind; }
-      long index_to_linear(index_t ind) const { return ind; }
+      long index_to_point(index_t idx) const {
+        EXPECTS(is_within_boundary(idx));
+        return idx;
+      }
+      long index_to_linear(index_t idx) const {
+        EXPECTS(is_within_boundary(idx));
+        return idx;
+      }
 
       // -------------------- mesh_point -------------------
 
@@ -70,12 +80,9 @@ namespace triqs {
 
       // -------------- Evaluation of a function on the grid --------------------------
 
-      /// Is the point in the mesh ?
-      bool is_within_boundary(index_t const &p) const { return ((p >= 0) && (p < size())); }
-      //bool is_within_boundary(index_t const &p) const { return ((p >= first_index_window()) && (p <= last_index_window())); }
-
       long get_interpolation_data(long n) const { return n; }
       template <typename F> auto evaluate(F const &f, long n) const { return f[n]; }
+
       // -------------------- MPI -------------------
 
       // -------------------- HDF5 -------------------
--- a/triqs/gfs/meshes/discrete.hpp
+++ b/triqs/gfs/meshes/discrete.hpp
@@ -52,7 +52,7 @@ namespace triqs {
       utility::mini_vector<size_t, 1> size_of_components() const { return {size()}; }
 
       /// Is the point in the mesh ?
-      bool is_within_boundary(all_t) const { return true; }
+      static constexpr bool is_within_boundary(all_t) { return true; }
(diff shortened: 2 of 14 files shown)


test/triqs/gfs/gf_empty.cpp
```diff
@@ -9,6 +9,6 @@ TEST(Gf, Exceptions) {
 
   auto g = gf<imtime, scalar_valued>{};
 
-  EXPECT_THROW(g[closest_mesh_pt(5.3)], triqs::arrays::key_error);
+  ASSERT_DEATH(g[closest_mesh_pt(5.3)], "");
```
parcollet, line 12: why this change ?

triqs/gfs/meshes/discrete.hpp
```diff
@@ -51,9 +51,19 @@ namespace triqs {
       ///
       utility::mini_vector<size_t, 1> size_of_components() const { return {size()}; }
 
+      /// Is the point in the mesh ?
+      bool is_within_boundary(all_t) const { return true; }
```
parcollet: by definition all_t is within mesh !
We should not have to implement this in **each** mesh, it should be intercepted
at the caller

triqs/gfs/meshes/imfreq.hpp
```diff
@@ -134,14 +134,26 @@ namespace triqs::gfs {
     ///
     utility::mini_vector<size_t, 1> size_of_components() const { return {size_t(size())}; }
 
+    /// Is the point in mesh ?
+    bool is_within_boundary(all_t) const { return true; }
```
parcollet: same

triqs/gfs/meshes/imtime.hpp
```diff
@@ -55,7 +55,9 @@ namespace triqs {
       gf_mesh(double beta, statistic_enum S, long n_time_slices) : gf_mesh({beta, S}, n_time_slices) {}
 
       /// For imtime the point is always in the mesh, since we use anti-periodicity or periodicity. Needed for cartesian product.
-      bool is_within_boundary(double x) const { return true; }
+      static constexpr bool is_within_boundary(all_t) { return true; }
```
parcollet: Really ? I thought we had removed the periodicity in evaluation ??
To be discussed.

triqs/gfs/meshes/linear.hpp
```diff
@@ -55,22 +66,22 @@ namespace triqs {
 
       utility::mini_vector<size_t, 1> size_of_components() const { return {size_t(size())}; }
 
+      /// Is the point in mesh ?
+      bool is_within_boundary(all_t) const { return true; }
```
parcollet: same

triqs/lattice/cluster_mesh.hpp
```diff
@@ -118,17 +118,21 @@ namespace triqs {
       using index_t        = utility::mini_vector<long, 3>;
       using linear_index_t = long;
 
+      /// Reduce index modulo to the lattice.
```
parcollet, line 121: What about the 
modulo_reduce
at the end of the file and used in the mesh_point ?

parcollet (approved):
two main points to discuss : 
- g(tau) periodicity
- modulo_reduce

test/triqs/gfs/gf_empty.cpp
Wentzell: This change is necessary because `g[closest_mesh_pt(5.3)]` will now immediately violate the precondition `EXPECTS(_mesh.is_within_boundary(arg));`. I.e. the test can no longer check for an exception but has to assert that the program terminates.

triqs/gfs/meshes/discrete.hpp
Wentzell: As agreed we delay this interception until after the 2.2 release.

triqs/lattice/cluster_mesh.hpp
Wentzell: I have removed the `modulo_reduce` as it achieves the same goal as `index_modulo`

triqs/gfs/meshes/imtime.hpp
Wentzell: This is in fact a bug! I have pushed a fix. We are now falling back to the `is_within_boundary` implementations of the `linear_mesh`

triqs/lattice/cluster_mesh.hpp
```diff
@@ -257,21 +254,17 @@ namespace triqs {
...
       double operator[](int d) const { return operator()(d); }
       friend std::ostream &operator<<(std::ostream &out, mesh_point const &x) { return out << (lattice_point)x; }
-      mesh_point operator-() const { return mesh_point{*m, m->modulo_reduce({-index()[0], -index()[1], -index()[2]})}; }
+      mesh_point operator-() const { return mesh_point{*m, m->index_modulo({-index()[0], -index()[1], -index()[2]})}; }
       mesh_t const &mesh() const { return *m; }
     };
 
     // --- impl
-    inline mesh_point<cluster_mesh> cluster_mesh::operator[](index_t i) const { return mesh_point<cluster_mesh>{*this, this->modulo_reduce(i)}; }
+    inline mesh_point<cluster_mesh> cluster_mesh::operator[](index_t i) const { return mesh_point<cluster_mesh>{*this, this->index_modulo(i)}; }
```
parcollet: why do we keep the modulo here ??

Wentzell: That is a good question. I just did the renaming in the file.
We should just use `i` here protect it with `EXPECTS(is_within_boundary(i))`. Do you agree?

PR #753, opened 2019-11-09
Fix bug in lattice tools dos for multi-orbital models

mzingl:
Dear all,
this pull request fixes a wrong orbital index in the calculation of the dos for multi-orbital tight-binding models. The total density of states was ok, however the orbital resolved one was wrong.

This is a tiny 1-line bugfix, thus should be directly merge-able.
Please also consider backporting it to 2.2.x. and 2.1.x.

Thanks, Manuel


Wentzell:
Dear @mzingl,

Thank you for spotting this!

PR #771, opened 2020-04-27
Serialize atomdiag to be able to bcast atomdiag containers

the-hampel:
- brought back h5 serialization option
- make AtomDiag python object serializable via h5
- added tests for mpi bcast functionalities of AtomDiag object

### DIFF
--- a/test/pytriqs/atom_diag/atom_diag_bcast.py
+++ b/test/pytriqs/atom_diag/atom_diag_bcast.py
@@ -1,6 +1,5 @@
 from pytriqs.operators import *
 from pytriqs.gf import *
-from pytriqs.archive import *
 from pytriqs.atom_diag import *
 from pytriqs.utility.comparison_tests import *
 import numpy as np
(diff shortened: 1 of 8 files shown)


test/pytriqs/atom_diag/atom_diag_bcast.py
```diff
@@ -0,0 +1,75 @@
...
+# HDF5
+if mpi.is_master_node():
+    with HDFArchive('atom_diag_bcast.h5', 'w') as ar:
+        ar['ad'] = ad
+
+
+ad_bcast = None
+if mpi.is_master_node():
+    with HDFArchive('atom_diag_bcast.h5', 'r') as ar:
+        ad_bcast = ar['ad']
```
Wentzell: @the-hampel Why do you use hdf5 here? Maybe you could just do

```python
if not mpi.is_master_node():
    ad = None
ad_bcast = mpi.bcast(ad)
```

instead?

the-hampel: Yes, sounds good. I fxed that in the new commit.

PR #781, opened 2020-08-05
[mc_generic] Properly catch exception and stop

parcollet:
- When an exception occurs on a node, properly stop all nodes
  at the end of the MC sweep, by reporting and throwing an exception on
  each node in C++.
- Update test to actually test in mpi mode, as Google Test.

### DIFF
--- a/c++/triqs/mc_tools/mc_generic.hpp
+++ b/c++/triqs/mc_tools/mc_generic.hpp
@@ -287,8 +287,8 @@ namespace triqs::mc_tools {
       }
 
       // final reporting
-      if (status == 1) std::cerr << "rank " << c.rank() << ": mc_generic stops because of stop_callback" << std::endl;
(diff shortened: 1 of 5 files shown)


c++/triqs/mc_tools/mc_generic.hpp
```diff
@@ -247,10 +278,19 @@ namespace triqs::mc_tools {
...
+      if (node_monitor) {
+        node_monitor->finalize_communications();
+        if (node_monitor->emergency_occured()) TRIQS_RUNTIME_ERROR << "mc_generic stops because the calculation raised an exception on one node";
+      }
+
       // final reporting
-      if (status == 1) report << "mc_generic stops because of stop_callback";
-      if (status == 2) report << "mc_generic stops because of a signal";
-      report << "\n" << std::endl;
+      if (status == 1) std::cerr << "rank " << c.rank() << ": mc_generic stops because of stop_callback" << std::endl;
```
parcollet: why cerr ? in case of call back (clock), it will report on ALL nodes !?

PR #783, opened 2020-08-10
Deprecate string indices, Remove deprecated parts of API

Wentzell:
Remove various deprecated parts of the Python and C++ API and adjust the ChangeLog.
Also, we now deprecate the use of string indices in the Green function.


parcollet (approved):
Cf slack comment.
No technical comment.

PR #784, opened 2020-08-11
Update install.rst with instructions for compiling using EasyBuild

rokzitko:


Wentzell:
Thank you for adding this @rokzitko !

PR #785, opened 2020-08-12
[gf.py] bugfix in order to access and set elements of rank1 Gfs

richtma93:
It is not possible to access and set single or multiple elements of a rank1 Gf in python (eq.: G[0] or G[:2]) .
Single arguments are assumed to be mesh points.
This commit should fix the issue.

### DIFF
--- a/python/triqs/gf/gf.py
+++ b/python/triqs/gf/gf.py
@@ -305,8 +305,9 @@ class Gf(metaclass=AddMethod):
 
         # Only one argument. Must be a mesh point
         if not isinstance(key, tuple):
-            assert isinstance(key, (MeshPoint, Idx))
-            return self.data[key.linear_index if isinstance(key, MeshPoint) else self._mesh.index_to_linear(key.idx)]
+            if isinstance(key, (MeshPoint, Idx)):
+                return self.data[key.linear_index if isinstance(key, MeshPoint) else self._mesh.index_to_linear(key.idx)]
+            else: key = (key,)
 
         # If all arguments are MeshPoint, we are slicing the mesh or evaluating
         if all(isinstance(x, (MeshPoint, Idx)) for x in key):
--- a/python/triqs/gf/gf.py
+++ b/python/triqs/gf/gf.py
@@ -303,10 +303,37 @@ class Gf(metaclass=AddMethod):
         if key == self._full_slice:
             return self
 
-        # Only one argument. Must be a mesh point
         if not isinstance(key, tuple):
-            assert isinstance(key, (MeshPoint, Idx))
-            return self.data[key.linear_index if isinstance(key, MeshPoint) else self._mesh.index_to_linear(key.idx)]
+            if isinstance(key, (MeshPoint, Idx)):
+                return self.data[key.linear_index if isinstance(key, MeshPoint) else self._mesh.index_to_linear(key.idx)]
+            else:
+                assert self.target_rank == 1, "wrong number of arguments. Expected %s, got %s"%(self.target_rank, 1)
+                # Assume empty indices (scalar_valued)
+                ind = GfIndices([])
+
+                # String access: transform the key into a list integers
+                if isinstance(key, str):
+                    assert self._indices, "Got string indices, but I have no indices to convert them !"
+                    key_lst = [self._indices.convert_index(key,1)] # convert returns a slice of len 1
+
+                # Slicing with ranges -> Adjust indices
+                elif isinstance(key, slice): 
+                    key_lst = [key]
+                    ind = GfIndices([ v[k]  for k,v in zip(key_lst, self._indices.data)])
+
+                # Integer access
+                elif isinstance(key, int):
+                    key_lst = [key]
+
+                # Invalid Access
+                else:
+                    raise NotImplementedError, "Partial slice of the target space not implemented"
+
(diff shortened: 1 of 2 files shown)


Wentzell:
Dear @richtma93,

Thank you for pointing out this issue and for drafting a solution!

It seems that your changes are breaking various Python tests however (c.f. [log](https://jenkins.flatironinstitute.org/job/TRIQS/job/triqs/view/change-requests/job/PR-785/2/execution/node/135/log/))
Could you look into this?

It would be nice if you could also add a small test to e.g. [test/python/base/gf_init.py](https://github.com/TRIQS/triqs/blob/unstable/test/python/base/gf_init.py)

richtma93:
Dear @Wentzell,

> It seems that your changes are breaking various Python tests however (c.f. [log](https://jenkins.flatironinstitute.org/job/TRIQS/job/triqs/view/change-requests/job/PR-785/2/execution/node/135/log/))
> Could you look into this?

Yes! However I think this is because I used the latest commit on the unstable branch ( 56c11cba36 ) to checkout this branch (rank1_compat). To my understanding the commit on the unstable branch ( 56c11cba36 ) already broke various Python tests.

I am honestly using the 2.2.x branch only and added the above changes to my local repository without any test failure.
Anyhow I thought that the target branch of a pull request should always be unstable?!

Wentzell:
> Yes! However I think this is because I used the latest commit on the unstable branch ( [56c11cb](https://github.com/TRIQS/triqs/commit/56c11cba3666d4095eab75691a2e9f5af70d6bca) ) to checkout this branch (rank1_compat). To my understanding the commit on the unstable branch ( [56c11cb](https://github.com/TRIQS/triqs/commit/56c11cba3666d4095eab75691a2e9f5af70d6bca) ) already broke various Python tests.

The latest commit on unstable is passing all tests. The branch is automatically built on Jenkins.
All the recent logs can be found here: https://jenkins.flatironinstitute.org/job/TRIQS/job/triqs/job/unstable/

Note however that TRIQS/unstable is now based based on Python 3 and not Python 2.

> I am honestly using the 2.2.x branch only and added the above changes to my local repository without any test failure.
> Anyhow I thought that the target branch of a pull request should always be unstable?!

It would be nice to have the fix available for 2.2.x, but it should definitely be added to unstable.
The usual procedure is to create the PR for the unstable branch and then, if it is an essential fix,
backport it to the latest release branch (2.2.x).

Let me know if you have any trouble updating the PR for unstable.

richtma93:
> Note however that TRIQS/unstable is now based based on Python 3 and not Python 2.

Yes turns out this was the problem. It should work now.
I also added a test to test/python/base/gf_init.py.

python/triqs/gf/gf.py
```diff
@@ -302,10 +302,37 @@ def __getitem__(self, key):
...
+                return self.data[key.linear_index if isinstance(key, MeshPoint) else self._mesh.index_to_linear(key.idx)]
+            else:
+                assert self.target_rank == 1, "wrong number of arguments. Expected %s, got %s"%(self.target_rank, 1)
+                # Assume empty indices (scalar_valued)
+                ind = GfIndices([])
+
+                # String access: transform the key into a list integers
+                if isinstance(key, str):
+                    assert self._indices, "Got string indices, but I have no indices to convert them !"
+                    key_lst = [self._indices.convert_index(key,1)] # convert returns a slice of len 1
```
richtma93: I am not sure if this is necessary, as I don't know any possibility of constructing a rank1 Gf with string indices...

python/triqs/gf/gf.py
```diff
@@ -302,10 +302,37 @@ def __getitem__(self, key):
...
+
+                # Invalid Access
+                else:
+                    raise NotImplementedError("Partial slice of the target space not implemented")
+
+                dat = self._data[ self._rank * (slice(0,None),) + key_tpl ] 
+                r = Gf(mesh = self._mesh, data = dat, indices = ind)
+
+                r.__check_invariants()
+                return r
```
Wentzell: This is a lot of duplication of the code that we already have in the else branch below.
Why don't you just change it in the following way:

- Revert this part to what it was before
- Change the condition on line 305 to `isinstance(key, (MeshPoint, Idx))`
- Inside the else branch below, the first thing you do is `if not isinstance(key, tuple): key = (key,)`

This should achieve the same goal and the logic is still easy to follow. What do you think?

Wentzell:
Thank you for the update! C.f. comment above.
Also can you rebase onto the latest triqs/unstable? The conflict should be easy to resolve.

python/triqs/gf/gf.py
richtma93: > This is a lot of duplication of the code that we already have in the else branch below.
> Why don't you just change it in the following way:
> 
>     * Revert this part to what it was before
> 
>     * Change the condition on line 305 to `isinstance(key, (MeshPoint, Idx))`
> 
>     * Inside the else branch below, the first thing you do is `if not isinstance(key, tuple): key = (key,)`
> 
> 
> This should achieve the same goal and the logic is still easy to follow. What do you think?

Sure this is an easier solution!
However I had to do the key = (key,) thing already before  'if all(isinstance(x, (MeshPoint, Idx)) for x in key):' to make sure that key is an iterable object.

Wentzell:
This was merged in c176dc458

Issue #719, opened 2019-06-24
Feature Request: Weighted Tailfit in 2.x


hschnait:
Dear All

It would be nice to have the possibility of a weighted tail-fit in 2.x. (as mentioned in PR #482 ).
This should work together nicely with the `fit_tail()` and `fit_hermitian_tail()` functions.

As the error of Sigma increases with w_n, these two functions mainly fit noise and are rather unusable for my application in their current implementation.

Wentzell:
Dear Hermann,

Thank you for the feature request! The PR you mention unfortunately got lost during the major
tail rework.

We agree that the capability for weighted tailfits is essential, in particular in view of
the on-going rework of the statistics functionality that is about to be merged into triqs/unstable.
We will let you know as soon as we start to move on this.

Best,

Nils

PR #789, opened 2020-08-31
Weighted Tailfit

hschnait:
Dear all

This PR implements the weighted tailfit originally proposed by @gkraberger (PR #482), that got lost during the rework of the tails in 2.x (see also #719).

Please review whether you want to add it in this way - or if I'm messing with something I should not mess with here.
The default behaviour of the tailfit does not change, the order of the error can be set by `set_tail_fit_parameters`.

### DIFF
--- a/c++/triqs/gfs/meshes/tail_fitter.hpp
+++ b/c++/triqs/gfs/meshes/tail_fitter.hpp
@@ -30,10 +30,10 @@ namespace triqs::gfs {
 
   //----------------------------------------------------------------------------------------------
   // construct the Vandermonde matrix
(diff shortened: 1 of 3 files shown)


the-hampel:
@Wentzell can we pick this up again and maybe merge? I thing this might be a very useful feature, because the default tail fit by fraction is rather useless at the moment without weighting by frequencies, as the tail fit is dominated by Gf elements from frequencies with very larger error.  I checked a bit the commits, and the changes seem to be rather small and do not break any backward compatibility if I see this correctly.

Wentzell:
Yes we should definitely pick this up again!

The current implementation is a bit stringent as it limits you to a particular analytic form for the fitting weights.
We should generalize and have the user provide a function that returns the weight based on Matsubara index.

PR #792, opened 2020-10-01
[atom_diag] Extend autopartition function with additionnal effective hyb to create new block structure.

mnmpdadish:


Wentzell:
This was merged into unstable.

PR #794, opened 2020-10-08
Test for issue when multiplying Gfs with different dimensions

the-hampel:
When multiplying two Green functions with different shape, for example for downfolding / upfolding purposes, triqs gives an error, that the dimensions do not match. A simple example would be: 
```
from triqs.gf import *

iw_mesh = MeshImFreq(beta=40, S='Fermion', n_max=1025)
# create upfolded Gf with 5x5 dim
G_upf = Gf(mesh=iw_mesh, target_shape=[5, 5])
# create downfolded target Gf with 3x3 dim
G_dnf = Gf(mesh=iw_mesh, target_shape=[3, 3])

# downfolding projector from the 5 to 3 orb
downfolding = Gf(mesh=iw_mesh, target_shape=[3,5])

# the downfolding / projection from 5 to 3 orbitals fails
# this should work since we are doing [3,5]x[5,5]x[5,3]
G_dnf  << downfolding*(G_upf*downfolding.transpose().conjugate())
```
This should work as the multiplication of matrices of these dimensions makes sense and is often used for downfolding. The code gives the error:
```
File "/mnt/home/ahampel/Dropbox/git/triqs3/triqs.src/test/python/base/issue_gf_mult_dim.py", line 36, in <module>
G_dnf  << downfolding*(G_upf*downfolding.transpose().conjugate())
File "/mnt/home/ahampel/git/triqs-3.x_unstable/triqs_fix_gf_mult.build/python/triqs/gf/gf.py", line 537, in __mul__
     c.__imul__impl(y)
File "/mnt/home/ahampel/git/triqs-3.x_unstable/triqs_fix_gf_mult.build/python/triqs/gf/gf.py", line 494, in __imul__impl
     d_self[n] = np.dot(d_self[n], d_args[n]) # put to C if too slow.
ValueError: could not broadcast input array from shape (5,3) into shape (5,5)
```
Triqs expect the outcoming Gf to have the same shape as the first input Gf.

This PR adds a simple test that fails at the moment and should be fixed.

Thanks.

### DIFF
--- a/python/triqs/gf/gf.py
+++ b/python/triqs/gf/gf.py
@@ -482,18 +482,6 @@ class Gf(metaclass=AddMethod):
         return c
 
     # ---------- Multiplication
-    # Naive implementation of G1 *= G2 without checks
-    def __imul__impl(self,arg): # need to separate for the ImTime case
-
-        # reshaping the data. Smash the mesh indices into one
-        rh = lambda d : np.reshape(d, (reduce(mul, d.shape[:self.rank]),) + (d.shape[self.rank:]))
-        d_self =  rh(self.data)
-        d_args =  rh(arg.data)
-        if self.target_rank == 2:
-            for n in range (d_self.shape[0]):
-               d_self[n] = np.dot(d_self[n], d_args[n]) # put to C if too slow.
-        else:
-            d_self *= d_args # put to C if too slow.
 
     def __imul__(self,arg):
         if descriptor_base.is_lazy(arg): return lazy_expressions.make_lazy(self) * arg
@@ -501,7 +489,24 @@ class Gf(metaclass=AddMethod):
         if isinstance(arg, Gf):
             assert type(self.mesh) == type(arg.mesh), "Can not multiply two Gf with meshes of different type"
             assert self.mesh == arg.mesh, "Can not use in-place multiplication for two Gf with different mesh"
-            self.__imul__impl(arg)
+
+            if self.target_rank == 2:
+                if arg.target_rank == 2:
+                    assert arg.target_shape[0] != arg.target_shape[1], "In place multiplication not supported if"
+                    "the argument is a nonsquare matrix. Use regular multiplication instead."
+                    np.matmul(self.data, arg.data, out=self.data)
+                elif arg.target_rank == 0:
+                    self.data[:] *= arg.data[..., None, None]
+                else:
+                    raise NotImplementedError("argument of in place multiplication must be rank 0 or 2")
+            
+            elif self.target_rank == 0:
+                assert arg.target_rank == 0, "argument of in place multiplication must have rank 0 if self does"
+                self.data[:] = self.data * arg.data
+            
+            else:
+                raise NotImplementedError("Green's functions must be of rank 0 or 2 for multiplication")
+
         elif isinstance(arg, numbers.Number):
             self._data[:] *= arg
         elif isinstance(arg, np.ndarray):
@@ -530,11 +535,23 @@ class Gf(metaclass=AddMethod):
     def __mul__(self,y):
         if isinstance(y, Gf):
             # make a copy, but special treatment of the mesh in the Imtime case.
-            c = Gf(mesh = Gf._combine_mesh_mul(self._mesh, y.mesh),
-                   data = self._data.copy(), 
-                   indices = self._indices.copy(), 
-                   name = self.name)
-            c.__imul__impl(y)
+            result_mesh = Gf._combine_mesh_mul(self._mesh, y.mesh)
+            if self.target_rank == 2 and y.target_rank == 2:
+                c = Gf(mesh=result_mesh, target_shape=[self.target_shape[0], y.target_shape[1]])
+                np.matmul(self.data, y.data, out=c.data)
+            elif self.target_rank == 0 and y.target_rank == 0:
+                c = Gf(mesh=result_mesh, target_shape=[])
+                c.data[:] = self.data * y.data
+            elif self.target_rank == 2 and y.target_rank == 0:
+                c = Gf(mesh=result_mesh, target_shape=self.target_shape)
+                c.data[:] = self.data * y.data[..., None, None]
+            elif self.target_rank == 0 and y.target_rank == 2:
+                c = Gf(mesh=result_mesh, target_shape=y.target_shape)
+                c.data[:] = self.data[..., None, None] * y.data
+
+            else:
+                raise NotImplementedError("Green's functions must be of rank 0 or 2 for multiplication")
+
         elif isinstance(y, (numbers.Number, np.ndarray)):
             c = self.copy()
             c *= y
--- a/test/python/base/gf_base_op.py
+++ b/test/python/base/gf_base_op.py
@@ -181,5 +181,23 @@ class test_Gf_Base_Op(unittest.TestCase):
         assert_gfs_are_close(G, G_exact)
         assert_gfs_are_close(Mat * G * linalg.inv(Mat), G_exact)
 
+    def test_different_rank_prod(self):
+        mesh = MeshImFreq(beta=self.beta, S="Fermion", n_max=10)
+        G1 = Gf(mesh=mesh, target_shape=[2,2])
+        G1 << inverse(iOmega_n + 2)
+        G2 = Gf(mesh=mesh, target_shape=[])
+        G2 << inverse(iOmega_n - 2)
+
+        G_loop = G1.copy()
+        for i in range(len(G1.mesh)):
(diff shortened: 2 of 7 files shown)


python/triqs/gf/gf.py
```diff
@@ -530,11 +530,21 @@ def _combine_mesh_mul(l, r):
...
-            c.__imul__impl(y)
+            if self.target_rank == 2 and y.target_rank == 2 and self.target_shape != y.target_shape:
+                c = Gf(mesh=Gf._combine_mesh_mul(self._mesh, y.mesh),
+                       target_shape=[self.target_shape[0], y.target_shape[1]])
+                rh = lambda d : np.reshape(d, (reduce(mul, d.shape[:c.rank]),) + (d.shape[c.rank:]))
+                d_self =  rh(self.data)
+                d_y =  rh(y.data)
+                d_c = rh(c.data)
+                for n in range (d_self.shape[0]):
+                    d_c[n] = np.dot(d_self[n], d_y[n])
```
Wentzell: Why all the reshapes? Shouldn't this just be
```suggestion
                np.dot(self.data, y.data, out=c.data)
```
c.f. https://numpy.org/doc/stable/reference/generated/numpy.dot.html
Quote:

> `numpy.dot(a, b, out=None)`
> If a is an N-D array and b is an M-D array (where M>=2), it is a sum product over the last axis of a and the second-to-last axis of b:
> `dot(a, b)[i,j,k,m] = sum(a[i,j,:] * b[k,:,m])`

Wentzell (changes requested):

jkarp314:
@Wentzell I think the PR should be ready now

python/triqs/gf/gf.py
```diff
@@ -530,11 +521,15 @@ def _combine_mesh_mul(l, r):
...
-                   name = self.name)
-            c.__imul__impl(y)
+            if self.target_rank == 2 and y.target_rank == 2 and self.target_shape != y.target_shape:
+                c = Gf(mesh=Gf._combine_mesh_mul(self._mesh, y.mesh),
+                       target_shape=[self.target_shape[0], y.target_shape[1]])
+            else:
+                c = Gf(mesh = Gf._combine_mesh_mul(self._mesh, y.mesh),
+                    data = self._data.copy(), 
+                    indices = self._indices.copy(), 
+                    name = self.name)
```
Wentzell:
```suggestion
            c = Gf(mesh=Gf._combine_mesh_mul(self._mesh, y.mesh),
                   target_shape=[self.target_shape[0], y.target_shape[1]])
```
I don't think we should make a distinction here. Note that we have deprecated the use of `self._indices` and the name of the Gf should not propagate through multiplication.

python/triqs/gf/gf.py
```diff
@@ -530,11 +521,15 @@ def _combine_mesh_mul(l, r):
...
+            np.matmul(self.data, y.data, out=c.data)
```
Wentzell: This actually requires `self.target_rank == 2` to make any sense. We should check this both here and in the `imul`.
Also, the `self.target_rank == 0` case (i.e. scalar product) should be properly handled and tested for both cases `mul` and `imul`.

Wentzell (changes requested):

python/triqs/gf/gf.py
```diff
@@ -482,26 +482,31 @@ def __rsub__(self,y):
...
         # If arg is a Gf
         if isinstance(arg, Gf):
             assert type(self.mesh) == type(arg.mesh), "Can not multiply two Gf with meshes of different type"
             assert self.mesh == arg.mesh, "Can not use in-place multiplication for two Gf with different mesh"
-            self.__imul__impl(arg)
+
+            if self.target_rank == 2:
+                if arg.target_rank == 2:
+                    assert arg.target_shape[0] != arg.target_shape[1], "In place multiplication not supported if"
+                    "the argument is a nonsaure matrix. Use regular multiplication instead."
```
Wentzell:
```suggestion
                    "the argument is a non-square matrix. Use regular multiplication instead."
```

python/triqs/gf/gf.py
```diff
@@ -482,26 +482,31 @@ def __rsub__(self,y):
...
+                    np.matmul(self.data, arg.data, out=self.data)
+                elif arg.target_rank == 0:
+                    self.data[:] = self.data*(np.broadcast_to(arg.data, self.data.T.shape).T)
```
Wentzell: Why not just:
```suggestion
                    self.data *= arg.data[..., None, None]
```

python/triqs/gf/gf.py
```diff
@@ -530,11 +535,26 @@ def _combine_mesh_mul(l, r):
     def __mul__(self,y):
         if isinstance(y, Gf):
             # make a copy, but special treatment of the mesh in the Imtime case.
-            c = Gf(mesh = Gf._combine_mesh_mul(self._mesh, y.mesh),
-                   data = self._data.copy(), 
-                   indices = self._indices.copy(), 
-                   name = self.name)
-            c.__imul__impl(y)
+            if self.target_rank == 2 and y.target_rank == 2:
+                c = Gf(mesh=Gf._combine_mesh_mul(self._mesh, y.mesh),
```
Wentzell: Define `result_mesh` using `_combine_mesh_mul` before going into the various cases?

python/triqs/gf/gf.py
```diff
@@ -530,11 +535,26 @@ def _combine_mesh_mul(l, r):
...
+                       target_shape=[self.target_shape[0], y.target_shape[1]])
+                np.matmul(self.data, y.data, out=c.data)
+            elif self.target_rank == 0 and y.target_rank == 0:
+                c = Gf(mesh=Gf._combine_mesh_mul(self._mesh, y.mesh),
+                       target_shape=[])
+                c.data[:] = self.data * y.data
+            elif self.target_rank == 2 and y.target_rank == 0:
+                c = Gf(mesh=Gf._combine_mesh_mul(self._mesh, y.mesh),
+                       target_shape=self.target_shape)
+                c.data[:] = self.data*(np.broadcast_to(y.data, self.data.T.shape).T)
```
Wentzell:
```suggestion
                c.data[:] = self.data * y.data[..., None, None]
```

python/triqs/gf/gf.py
```diff
@@ -530,11 +535,26 @@ def _combine_mesh_mul(l, r):
...
+            elif self.target_rank == 0 and y.target_rank == 2:
+                c = Gf(mesh=Gf._combine_mesh_mul(self._mesh, y.mesh),
+                       target_shape=y.target_shape)
+                c.data[:] = y.data*(np.broadcast_to(self.data, y.data.T.shape).T)
```
Wentzell:
```suggestion
                c.data[:] = self.data[..., None, None] * y.data
```

test/python/base/gf_block.py
```diff
@@ -88,5 +88,43 @@ def test_block2_real_imag(self):
...
+        G2 << inverse(2 - iOmega_n)
+        for i in range(2):
+            G3[0,i] << inverse(iOmega_n + 2 + i)
+        for i in range(3):
+            G4[0,i] << inverse(iOmega_n + 2 + i)
+
+        B1 = BlockGf(name_list=['0', '1'], block_list=[G1, G2])
+        B2 = BlockGf(name_list=['0', '1'], block_list=[G3, G4])
+        B3 = B1 * B2
+        B4 = BlockGf(name_list=['0', '1'], block_list=[G1*G3, G2*G4])
```
Wentzell, line 108:
```suggestion
        B4 = BlockGf(name_list=['0', '1'], block_list=[G1*G3, G2*G4])
        
        assert_block2_gfs_are_close(B3, B4)
```

test/python/base/gf_base_op.py
```diff
@@ -181,5 +181,19 @@ def test_Mat_Prod(self):
...
+        G1 << inverse(iOmega_n + 2)
+        G2 = Gf(mesh=mesh, target_shape=[])
+        G2 << inverse(iOmega_n - 2)
+
+        G_mult = G1.copy()
+        G_mult << G1 * G2
+        G_loop = G1.copy()
+        for i in range(len(G1.mesh)):
+            G_loop.data[i] = G1.data[i] * G2.data[i]
+        assert_gfs_are_close(G_mult, G_loop)
```
Wentzell: Maybe add comparison also with exact result, i.e. `G3 << inverse(iOmega_n + 2) * inverse(iOmega_n - 2)`

Wentzell (changes requested):

Wentzell:
Thank you @jkarp314 and @the-hampel

PR #799, opened 2020-11-18
Add missing #pragma once and includes in utility/variant_extensions.hpp

krivenko:


Wentzell:
Thank you @krivenko !

PR #800, opened 2021-01-26
tests: Improve error messages and fix failure with GCC 10

hmenke:
Currently the unit tests only use the `assertTrue` method of the `unittest` framework.  This results in useless error messages à la `False is not true` in the case of a test failure.  By using the [provided assertion methods](https://docs.python.org/3/library/unittest.html#assert-methods) these messages can be improved substantially.

This came up because I compiled TRIQS with GCC 10 (in GitHub Actions) and ran the tests where the test `py_tail_issues` failed repeatedly.
```
  File "/build/source/test/python/base/tail_issues.py", line 137, in test_delta_infty
    self.assertLess(max_im, 1e-12)
AssertionError: 1.8302364448227348e-12 not less than 1e-12
```
To this end I propose increasing the test tolerance from `1e-12` to `2e-12`.

### DIFF
--- a/test/python/base/tail_issues.py
+++ b/test/python/base/tail_issues.py
@@ -36,7 +36,7 @@ class test_tail_issues(unittest.TestCase):
 
         # Check that tail fit error exceeds fourier error boundrary of 1e-2
         tail, err = g.fit_tail()
-        self.assertTrue(err > 1e-2)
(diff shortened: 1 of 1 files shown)


Wentzell:
Looks good! Thank you @hmenke

PR #801, opened 2021-01-27
[doc] Replace remaining RI by mode

hmenke:
The RI key for oplot was deprecated in 37e68715 and finally removed in 18fc34f3 but the documentation and examples were not updated in all places.


Wentzell:
Thank you for these doc improvements @hmenke

PR #802, opened 2021-02-05
add discretize bath function for Delta_iw objects

the-hampel:
* added function discretize_bath for scalar values Gf like Delta_iw
* uses scipy minimize Nelder-Mead method to optimizie input hopping and
  energy values to given tolerance [default=1e-08]
* added test for discretize_bath

### DIFF
--- a/python/triqs/gf/tools.py
+++ b/python/triqs/gf/tools.py
@@ -24,7 +24,7 @@ from .block_gf import BlockGf
 from .gf import Gf
 import numpy as np
 from itertools import product
-from .backwd_compat.gf_refreq import GfReFreq 
+from .backwd_compat.gf_refreq import GfReFreq
 from .map_block import map_block
 
 def inverse(x):
@@ -85,7 +85,7 @@ def delta(g):
         if err > 1e-5: print("WARNING: delta extraction encountered a sizeable tail-fit error: ", err)
         return delta_iw
     else:
-        raise TypeError("No function delta for g0 object of type %s"%type(g)) 
+        raise TypeError("No function delta for g0 object of type %s"%type(g))
 
 
 # Determine one of G0_iw, G_iw and Sigma_iw from other two using Dyson's equation
@@ -117,7 +117,7 @@ def dyson(**kwargs):
 def read_gf_from_txt(block_txtfiles, block_name):
     """
     Read a GfReFreq from text files with the format (w, Re(G), Im(G)) for a single block.
-    
+
     Notes
     -----
     A BlockGf must be constructed from multiple GfReFreq objects if desired.
@@ -152,7 +152,7 @@ def read_gf_from_txt(block_txtfiles, block_name):
 def write_gf_to_txt(g):
     """
     Write a GfReFreq or GfImFreq to in the format (w/iw, Re(G), Im(G)) for a single block.
-    
+
     Parameters
     ----------
     g: GfReFreq or GfImFreq
@@ -175,7 +175,7 @@ def write_gf_to_txt(g):
 def make_zero_tail(g, n_moments=10):
     """
     Return a container for the high-frequency coefficients of a given Green function initialized to zero.
-    
+
     Parameters
     ----------
     g: GfImFreq or GfReFreq or GfImTime or GfReTime
@@ -199,7 +199,7 @@ def fit_legendre(g_t, order=10):
     Only Hermiticity is imposed on the fit, so discontinuities has
     to be fixed separately (see the method enforce_discontinuity)
 
-    Author: Hugo U.R. Strand 
+    Author: Hugo U.R. Strand
 
     Parameters
     ----------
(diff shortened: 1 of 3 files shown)


python/triqs/gf/tools.py
```diff
@@ -259,8 +259,94 @@ def fit_legendre(g_t, order=10):
...
     scale = scale.reshape([len(lmesh)] + [1]*len(g_t.target_shape))
-    
+
     g_l = Gf(mesh=lmesh, target_shape=g_t.target_shape)
     g_l.data[:] = c_l.reshape(g_l.data.shape) / scale
 
     return g_l
+
+
+def discretize_bath(delta_inp, Nb , bandwidth = 3, t_val = 0.0, tol=1e-5):
```
Wentzell: Use delta_iw instead of delta_imp throughout?
```suggestion
def discretize_bath(delta_iw, Nb, bandwidth = 3, t_init = 0.0, tol=1e-5):
```

python/triqs/gf/tools.py
```diff
@@ -259,8 +259,94 @@ def fit_legendre(g_t, order=10):
...
+        number of bath sites
+    bandwidth : approximative bandwidth for init of bath energies
+    t_val : float
+        to which value should the bath hoppings be initialized
+    tol: float
+        tolerance for scipy minimize
+    Returns
+    -------
+    t_opt : list of optimized bath hopping parameters
+    e_opt : list of optimized bath energies
```
Wentzell:
```suggestion
    e_opt : sorted list of optimized bath energies
```

python/triqs/gf/tools.py
```diff
@@ -259,8 +259,94 @@ def fit_legendre(g_t, order=10):
...
+               Hybridization function.
+    """
+    from scipy.optimize import minimize
+
+    ####
+    # define minimizer for scipy
+    def minimizer(parameters):
+        # first half of parameters are hoppings
+        t = parameters[0:Nb]
+        # scond half are bath energies
```
Wentzell:
```suggestion
        # second half are bath energies
```

python/triqs/gf/tools.py
```diff
@@ -259,8 +259,94 @@ def fit_legendre(g_t, order=10):
...
+    """
+    discretizes a given Delta_iw with Nb bath sites
+
+    CAUTION: This function works only for 1x1 / scalar valued Gf objects.
+    if necessary call for each orbital in input Delta_iw separately.
+
+    Parameters
+    ----------
+    g : GfImFreq
+        Non-interacting Green's function.
```
Wentzell:
```suggestion
    delta_iw : Gf
        Matsubara hybridization function to discretize
```

python/triqs/gf/tools.py
```diff
@@ -259,8 +259,94 @@ def fit_legendre(g_t, order=10):
...
+
     g_l = Gf(mesh=lmesh, target_shape=g_t.target_shape)
     g_l.data[:] = c_l.reshape(g_l.data.shape) / scale
 
     return g_l
+
+
+def discretize_bath(delta_inp, Nb , bandwidth = 3, t_val = 0.0, tol=1e-5):
+    """
+    discretizes a given Delta_iw with Nb bath sites
```
Wentzell: Provide here the analytic form of the discretized hybridization function and the norm used for the optimization.
Mention also 'scipy.optimize.minimize'.

python/triqs/gf/tools.py
```diff
@@ -259,8 +259,94 @@ def fit_legendre(g_t, order=10):
...
+    ----------
+    g : GfImFreq
+        Non-interacting Green's function.
+    Nb : int
+        number of bath sites
+    bandwidth : approximative bandwidth for init of bath energies
+    t_val : float
+        to which value should the bath hoppings be initialized
+    tol: float
+        tolerance for scipy minimize
```
Wentzell:
```suggestion
    bandwidth : float, optional
        approximate bandwidth, used in choice of initial of bath energies [default=3.0]
    t_init : float, optional
        initial guess used for all hopping values [default=0.0]
    tol: float, optional
        tolerance for scipy minimize [default=1e-5]
```

Also we may want to use a more conservative default for the tolerance (~1e-8)?

python/triqs/gf/tools.py
```diff
@@ -259,8 +259,94 @@ def fit_legendre(g_t, order=10):
...
+    def minimizer(parameters):
+        # first half of parameters are hoppings
+        t = parameters[0:Nb]
+        # scond half are bath energies
+        e = parameters[Nb:]
+
+        # build discretized bath function as
+        # delta = sum_nb t_n / (iw - eps_n)
+        for w in delta_disc.mesh:
+            delta_disc[w] = np.sum(t**2 / ( w.value - e ) ) - delta_inp[w]
```
Wentzell:
```suggestion
            delta_disc[w] = np.sum(t**2 / ( w.value - e ) )
```

python/triqs/gf/tools.py
```diff
@@ -259,8 +259,94 @@ def fit_legendre(g_t, order=10):
...
+    t_val : float
+        to which value should the bath hoppings be initialized
+    tol: float
+        tolerance for scipy minimize
+    Returns
+    -------
+    t_opt : list of optimized bath hopping parameters
+    e_opt : list of optimized bath energies
+    delta_iw : BlockGf (of GfImFreq) or GfImFreq
+               Hybridization function.
```
Wentzell:
```suggestion
    delta_disc_iw : Gf
               Discretized Matsubara hybridization function
```

python/triqs/gf/tools.py
```diff
@@ -259,8 +259,94 @@ def fit_legendre(g_t, order=10):
...
+        # scond half are bath energies
+        e = parameters[Nb:]
+
+        # build discretized bath function as
+        # delta = sum_nb t_n / (iw - eps_n)
+        for w in delta_disc.mesh:
+            delta_disc[w] = np.sum(t**2 / ( w.value - e ) ) - delta_inp[w]
+
+        # calculate norm sum_iw_n | delta_disc - delta_inp |^2
+        norm = (delta_disc*delta_disc.conjugate()).density().real
```
Wentzell: I would not use density for the norm as this will trigger a tailfit of the possibly noisy data
```suggestion
        norm = np.linalg.norm(delta_disc.data - delta_iw.data)
```

python/triqs/gf/tools.py
```diff
@@ -259,8 +259,94 @@ def fit_legendre(g_t, order=10):
...
+    ####
+
+    # some tests if input is okay
+    assert not isinstance(delta_inp, BlockGf), 'delta_inp needs to be a GfImFreq not BlockGf'
+    assert isinstance(delta_inp.mesh, MeshImFreq), 'input delta_iw should have a mesh MeshImFreq'
+    assert len(delta_inp.target_shape)==0, 'delta_inp should have rank 0, i.e. a scalar valued Green function'
+
+    # prepare discretized delta with same mesh
+    delta_disc = delta_inp.copy()
+    delta_disc.zero()
```
Wentzell: Move lines 315-322 before definition of minimizer function?
This way delta_disc is defined before it first appears in the minimizer definition.

python/triqs/gf/tools.py
```diff
@@ -259,8 +259,94 @@ def fit_legendre(g_t, order=10):
...
+    t_opt = np.abs(result.x[0:Nb])
+
+    # sort by energy
+    order = np.argsort(e_opt)
+    e_opt = e_opt[order]
+    t_opt = t_opt[order]
+
+    # create optimized discretized delta on same mesh
+    for w in delta_disc.mesh:
+        delta_disc[w] = np.sum(t_opt**2/( w.value - e_opt ) )
```
Wentzell:
```suggestion
```
You can skip this step as the minimizer will already have set delta_disc accordingly

Wentzell (changes requested):
Thank you @the-hampel for this useful addition!
See my comments below

python/triqs/gf/tools.py
the-hampel: I checked, and scipy optimize as problems for that small tolerances. Sometimes 1e-8 will fail. We could maybe go for 1e-7 or 1e-6 for now. Will change that.

python/triqs/gf/tools.py
the-hampel: Edit: nevermind, found the problem. Works!
Hm. I thought so too, but in my example it did not work. delta_disc looks quite different if I construct it explicitly again from the parameters provided by the function. Maybe I did something wrong? 

image

python/triqs/gf/tools.py
the-hampel: did that see 14b57cc

python/triqs/gf/tools.py
the-hampel: did that

Wentzell (approved):

PR #806, opened 2021-02-24
fix indents in triqs sumk

the-hampel:
Small fix for triqs sumk lattice tools. Idention levels were not correct. Before a simple 
```
from triqs.sumk import *
``` 
did not work. Fixed now.


Wentzell:
Thank you @the-hampel

PR #808, opened 2021-03-17
update to discretize bath function

the-hampel:
- proper error message from scipy minimize is now displayed
- passing through param maxiter to scipy minimize


Wentzell:
Thank you @the-hampel

PR #809, opened 2021-03-19
Run triqs without MPI

the-hampel:
Adding the functionality to all triqs parts to be run without calling any parts of MPI. Even not an init. This allows to run the c++ and python part to be run without having access to MPI at runtime for testing, or quick analysis. Can be useful on HPC systems with no MPI access on login nodes. This needs also an updated version of triqs_mpi and NDA (PR seperately).

- added a second mpi layer which is imported instead of mpi4py if triqs is run without mpirun
- added a test to check if the mpi environment can be detected correctly
- added tests for running w/o MPI
- changed parts of c++ code to be in line with triqs_mpi

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -106,16 +106,12 @@ set(CMAKE_CXX_EXTENSIONS OFF)
 
 cmake_policy(SET CMP0069 NEW)
 include(CheckIPOSupported)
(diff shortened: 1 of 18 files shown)


parcollet:
Reviewed, it looks good to me. Merge when ready, if pass all tests.

PR #816, opened 2021-07-08
fix Wreturn-type warning for non void function without return type

the-hampel:


Wentzell:
Thank you @the-hampel

PR #817, opened 2021-08-19
update ubuntu in docker to 21.04 to resolve sphinx comp issues

the-hampel:
note:
- libunwind has no version in official repos (cthyb doc build works without problems. Nbsphinx works)
- ipython needs to be installed for nbconvert to work in 21.04


Wentzell:
Thank you @the-hampel !

PR #818, opened 2021-08-20
fix C order problems in SumK

the-hampel:
Not urgent, but this is a simple fix for problems with numpy arrays being transposed but not copied before giving to NDA.


Wentzell:
Thank you @the-hampel

Issue #819, opened 2021-08-30
[hilbert_space] Serious bug in `imperative_operator` introduced by #686


krivenko:
Hi everyone,

My collaborators noticed a strange behavior of `atom_diag`, when energy levels of the same Hamiltonian were different depending on the order of index combinations in the supplied `fops` list.

The following minimal example demonstrates the problem.
```Python
from triqs.operators import c, c_dag, Operator
from triqs.atom_diag import *
from itertools import product

orbs = (1, 2, 3)
spins = ("dn", "up")

# Construct a sum of the pair hopping and spin flip terms from a 3-orbital Kanamori interaction Hamiltonian
H_p = Operator()
H_J = Operator()

for o1, o2 in product(orbs, repeat = 2):
    if o1 == o2: continue
    
    H_p += c_dag(o1, "dn") * c_dag(o1, "up") * c(o2, "up") * c(o2, "dn")
    H_J += c_dag(o1, "dn") * c(o1, "up") * c_dag(o2, "up") * c(o2, "dn")

H = H_p + H_J

fops1 = [(o, s) for (o, s) in product(orbs, spins)]
fops2 = [(o, s) for (s, o) in product(spins, orbs)]

ad1 = AtomDiag(H, fops1)
ad2 = AtomDiag(H, fops2)

for e1, e2 in zip(ad1.energies, ad2.energies):
    print(e1.round(5), e2.round(5))
```
The output I am getting on my `3.0.x` installation is as follows.
```
[0. 0. 3.] [0. 3. 3.]
[0. 2.] [0. 3. 3.]
[0. 2.] [1. 1. 4.]
[0. 2.] [1. 3.]
[0. 2.] [1. 3.]
[0. 2.] [1. 3.]
[0. 0. 3.] [1. 3.]
[0. 2.] [1. 3.]
[0. 2.] [1. 3.]
[0. 0. 3.] [1. 3.]
[0. 2.] [1. 3.]
[0. 0. 3.] [1. 1. 4.]
[0. 2.] [1. 3.]
[0. 2.] [1. 3.]
[0. 2.] [1. 3.]
[0. 2.] [1. 3.]
[1.] [2.]
[1.] [2.]
[1.] [2.]
...
[1.] [2.]
```
The energy levels are clearly different (the `E = 4.0` level is missing from `ad1`'s output).

[...]

PR #820, opened 2021-08-30
Fix a serious bug in `imperative_operator` (#819) introduced by PR #686

krivenko:
PR #686 changed the order in which `fundamental_operator_set` stores index combinations. Before that PR, mutual order of two index combinations was dictated by the result of `operator<()`, which had been consistent with the way `many_body_operator` orders canonical operators within a monomial. Now the order of the index combinations and - consequently - of single particle bits forming a Fock state is arbitrary. When computing fermionic sign prefactor, the constructor of `imperative_operator` relied on ordering of the bits being consistent with the ordering of the monomial. PR #686 broke that assumption, which resulted in a dangerous and hard-to-find bug -- sign of some matrix elements can be erroneously flipped.

This PR fixes the issue by sorting monomials in `imperative_operator`'s constructor and revealing a possible sign change due to a permutation of canonical operators caused by `fops`.

With this fix, the Python script from #819 gives the following output.
```
[0. 0. 3.] [0. 0. 3.]
[0. 2.] [0. 2.]
[0. 2.] [0. 2.]
[0. 2.] [0. 2.]
[0. 2.] [0. 2.]
[0. 2.] [0. 0. 3.]
[0. 0. 3.] [0. 2.]
[0. 2.] [0. 2.]
[0. 2.] [0. 2.]
[0. 0. 3.] [0. 2.]
[0. 2.] [0. 0. 3.]
[0. 0. 3.] [0. 0. 3.]
[0. 2.] [0. 2.]
[0. 2.] [0. 2.]
[0. 2.] [0. 2.]
[0. 2.] [0. 2.]
[1.] [1.]
[1.] [1.]
[1.] [1.]
...
[1.] [1.]
```

Despite the fact the base branch is set to `3.0.x` here, I would greatly appreciate having the fix on `2.2.x` as well.

### DIFF
--- a/c++/triqs/hilbert_space/imperative_operator.hpp
+++ b/c++/triqs/hilbert_space/imperative_operator.hpp
@@ -99,12 +99,38 @@ namespace triqs {
         hilbert_map = hmap;
         if ((hilbert_map.size() == 0) != !UseMap) TRIQS_RUNTIME_ERROR << "Internal error";
 
+        auto greater = [&fops](triqs::operators::canonical_ops_t const& op1,
+                               triqs::operators::canonical_ops_t const& op2) {
+          if(op1.dagger != op2.dagger) return op2.dagger;
+          return op1.dagger ? (fops[op1.indices] > fops[op2.indices]) :
+                              (fops[op1.indices] < fops[op2.indices]);
+        };
+
         // The goal here is to have a transcription of the many_body_operator in terms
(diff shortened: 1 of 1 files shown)


Wentzell:
Thank you @krivenko for this fix!

Wentzell:
This was merged into 2.2.x in babab7d8 and into unstable in fee87d74

PR #823, opened 2021-10-08
Change behavior of make_real_in_tau and is_gf_real_in_tau for matrix_valued

Wentzell:
These functions  were previously identical to `make_hermitian` and `is_gf_hermitian`,
while they should really be based on the Green Function relation `G_(ij..)[iw] = G_(ij..)[-iw]*`
which is associated with a purely real-valued G_(ij..)(tau) (as opposed to `G_ij[iw] = G_ji[-iw]*` for hermiticity).
This PR implements these changes and extends the tests for the same functions.


the-hampel (commented):
Looks good! This is a nice improvement. The implementation looks good. I will test the python implementation of the function briefly. But I think this can be merged as it is.

PR #824, opened 2021-10-14
expose atom_diag constructor for limiting particle subspaces

the-hampel:
* The constructor for diagonalizing only specific particle number subspaces was not yet exposed to python.
* fix python BrZone and CycLad mesh desc: Gf function with mesh BrZone and CycLad could not properly be accessed
  in python
* fix numpy deprecation warning in gf


Wentzell:
Thank you @the-hampel !

PR #826, opened 2021-11-11
BUG: use Sequence from collections.abc

DerWeh:
TRIQS uses `collections.Sequence` which does not exist.
I assume you wanted the abstract base class `collections.abc.Sequence`?

I changed import order in the fashing of [PEP8](https://www.python.org/dev/peps/pep-0008/#imports) using `impsort`


Wentzell:
Thank you @DerWeh !

PR #830, opened 2022-01-03
fix error in TBSuperLattice __repr__function

jkarp314:
The __repr__ function of TBSuperLattice currently mixes up "Remove internal Hoppings" and "Cluster site positions" when printing the information about the superlattice. 

I changed the order to fix it


Wentzell:
Thank you @jkarp314 !

PR #832, opened 2022-01-14
fix SumkDiscreteFromLattice after TBL changes

the-hampel:
after recent changes in triqs lattice utils Sumk_discrete_from_lattice.py has not been updated. 

* add numpy and fix usage
* exchange `hopping` -> `fourier`
* `SL.dim` is not working anymore, TBL has no property dim anymore, instead use correctly `SL.bl.ndim`

### DIFF
--- a/python/triqs/sumk/sumk_discrete_from_lattice.py
+++ b/python/triqs/sumk/sumk_discrete_from_lattice.py
@@ -19,6 +19,7 @@
 
 
 from .sumk_discrete import SumkDiscrete
+import numpy as np
 from triqs.lattice.tight_binding import TBLattice
 
 class SumkDiscreteFromLattice (SumkDiscrete):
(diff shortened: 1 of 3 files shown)


the-hampel:
todo:
- check if other things are not fixed yet. 
- maybe add a test after correct fix has been implemented.

python/triqs/sumk/sumk_discrete_from_lattice.py
```diff
@@ -46,7 +47,7 @@ def __init__(self, lattice, patch = None, n_points = 8, method = "Riemann"):
         self.SL = lattice
         self.patch,self.method = patch,method
         # init the array
-        SumkDiscrete.__init__ (self, dim = self.SL.dim, gf_struct = lattice.OrbitalNames)
+        SumkDiscrete.__init__ (self, dim = self.SL.bl.ndim, gf_struct = lattice.OrbitalNames)
```
Wentzell:
```suggestion
        SumkDiscrete.__init__ (self, dim = self.SL.ndim, gf_struct = lattice.OrbitalNames)
```

Wentzell: Thank you for the fix! This class did indeed go unnoticed in the `TBLattice` changes.
I only noticed the simplification above.

A test would certainly be great to have for the `SumkDiscreteFromLattice`!

the-hampel:
I added a very simple test for `sumk_discrete_from_lattice` and also fixed some problems in the class itself.

Wentzell:
Looks good! Thank you @the-hampel

Issue #825, opened 2021-10-30
triqs::atom_diag::quantum_number_eigenvalues returns incorrect values when constraining the number of particles


YannIntVeld:
### Description

We have been using the atom_diag class in Python to diagonalize small Hubbard systems. As we are only interested in fixed occupation numbers, we constrained the diagonalization to certain particle number subspaces. (By modifying the installation slightly as proposed by @the-hampel ) However, we find that the 'quantum_number_eigenvalues' functions no longer return correct information.

### Steps to Reproduce

First we modified the TRIQS installation slightly in order to be able to constrain atom_diag to fixed particle numbers in Python.
In https://github.com/TRIQS/triqs/blob/unstable/python/triqs/atom_diag/atom_diag_desc.py we added a new method call wrapper in line 64
```c++
c.add_constructor("(many_body_operator h, fundamental_operator_set fops, int n_min, int n_max)", doc = "Reduce a given Hamiltonian to a block diagonal, diagonalize only within particle numbers given")
```
and recompiled TRIQS.

Next we used a simple two-site tight-binding model to test our results. Our minimal code example:
```python
from triqs.operators import *
from triqs.operators.util.observables import *
from triqs.atom_diag import *
import numpy as np
from itertools import product

def get_multiplet_occ(atom_diag, orb_names, spin_names = ['up','down'], off_diag = True):
    '''
    function to extract operator expectation values from measured 
    density matrix and atom_diag object from cthyb solver.
    
    measures N, and constructs eigenstates, and probabilities
    
    Parameters
    ----------
    atom_diag: triqs atom_diag object
        contains information about the local Hamiltonian (Hloc_0 + H_int)
    orb_names : list of int
        list of orbital indices
    spin_names : list of string
        list of strings containing the spin channel names
    off_diag: boolean
        determines wether blocks of Gf are named up_0 (false) or just up (true)
    __Returns:__
    res : Panda DataFrame
        containing all results structured
    '''
    import pandas as pd
    
    
    # res will be a list of dictionaries to be a panda data frame
    res = []
    
    # get fundamental operators from atom_diag object
    occ_operators = []
    for op in atom_diag.fops:
        occ_operators.append( n(op[0], op[1]) )
    
    # construct total occupation operator from list
    N_op = sum(occ_operators)
        
    # get particle numbers from atom_diag
    particle_numbers = quantum_number_eigenvalues(N_op, atom_diag)
    print(particle_numbers)
[...]

PR #835, opened 2022-02-14
 [atom_diag] Another go at fixing #825

krivenko:
This solution is based on a conceptionally different implementation of `atom_diag::get_matrix_element_of_monomial()`. It is proposed by @Wentzell in https://github.com/TRIQS/triqs/issues/825#issuecomment-1036647227.

I have also made `atom_diag::get_matrix_element_of_monomial()` public and removed the free function `matrix_element_of_monomial()` from `atom_diag/impl/functions.cpp` as it was doing the very same thing.

### DIFF
--- a/c++/triqs/atom_diag/impl/atom_diag.cpp
+++ b/c++/triqs/atom_diag/impl/atom_diag.cpp
@@ -22,6 +22,8 @@
 #include "./worker.hpp"
 
 #include <triqs/arrays.hpp>
+#include <triqs/hilbert_space/state.hpp>
+#include <triqs/hilbert_space/imperative_operator.hpp>
 
 using namespace triqs::arrays;
 
@@ -47,7 +49,6 @@ namespace triqs {
       fill_first_eigenstate_of_subspace();
       compute_vacuum();
     }
-    
 
     // -----------------------------------------------------------------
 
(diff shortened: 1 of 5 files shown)


c++/triqs/atom_diag/impl/atom_diag.cpp
```diff
@@ -107,17 +108,48 @@ namespace triqs {
...
+      state<class hilbert_space, scalar_t, true> final_st(full_hs);
+
+      matrix_t m;
+      int Bp = -1;
+      bool found_nonzero_element = false;
+
+      auto const& B_fock_states = get_fock_states(B);
+      for(fock_state_t i_index = 0; i_index < B_fock_states.size(); ++i_index) {
+        fock_state_t i = B_fock_states[i_index];
+        initial_st(i) = scalar_t(1.0);
```
Wentzell: You are calling the `value_type & state<..>::operator()(int i) { return ampli[i]; }` here with a `fock_state_t` which is a `uint64_t`. Shouldn't `ampli` really be a `map<fock_state_t, ScalarType>` and the  `operator()(int)` should be a `operator()(fock_state_t)`?

Wentzell: How about adding a `state<..>` constructor from a `fock_state_t` and the `full_hs`? Then you could move the construction of `initial_st` into the loop and can skip zeroing it out. I wouldn't expect much performance difference as the `state` is already sparse and you have a single map insertion in both cases?

c++/triqs/atom_diag/impl/atom_diag.cpp
```diff
@@ -107,17 +108,48 @@ namespace triqs {
...
-        B = Bp;
+      state<class hilbert_space, scalar_t, true> initial_st(full_hs);
+      state<class hilbert_space, scalar_t, true> final_st(full_hs);
+
+      matrix_t m;
+      int Bp = -1;
+      bool found_nonzero_element = false;
+
+      auto const& B_fock_states = get_fock_states(B);
+      for(fock_state_t i_index = 0; i_index < B_fock_states.size(); ++i_index) {
```
Wentzell: `i_index` is not really a `fock_state_t` but rather the index of the fock_state?
You could do
```cpp
for(auto [fs_idx, fs]: itertools::enumerate(B_fock_states)) {
  ...
}

File: c++/triqs/atom_diag/impl/atom_diag.cpp [hunk 3]
```diff
@@ -107,17 +108,48 @@ namespace triqs {
...
+
+              if(not found_nonzero_element) {
+                m.resize(get_subspace_dim(Bp), get_subspace_dim(B));
+                m() = scalar_t(.0);
+                found_nonzero_element = true;
+              }
+
+              m(sp.get_state_index(j), i_index) = x;
+              continue;
+            }
```
Wentzell: We can break out of the subspace loop as soon as we have found a single non-zero element?
There should be exactly one such element.
```suggestion
            if(sp.has_state(j) and std::abs(x) > 0.0) {
              Bp = sp.get_index();
              m = matrix_t::zeros({get_subspace_dim(Bp), get_subspace_dim(B)});
              m(sp.get_state_index(j), i_index) = x;
              break;
            }
```

c++/triqs/atom_diag/impl/atom_diag.cpp
```diff
@@ -107,17 +108,48 @@ namespace triqs {
...
+      matrix_t m;
+      int Bp = -1;
+      bool found_nonzero_element = false;
+
+      auto const& B_fock_states = get_fock_states(B);
+      for(fock_state_t i_index = 0; i_index < B_fock_states.size(); ++i_index) {
+        fock_state_t i = B_fock_states[i_index];
+        initial_st(i) = scalar_t(1.0);
+
+        final_st = monomial_op(initial_st);
```
Wentzell: Can we check that there is exactly one entry in the `final_st` ?

Wentzell (commented):
Dear @krivenko,

Thank you for drafting this fix!
I added a few comments below. Please let me know what you think.

c++/triqs/atom_diag/impl/atom_diag.cpp
krivenko: As of now, the map-based version of `state` does not give access to the amplitude container or its size. Do you think it is worth adding a new accessor?

c++/triqs/atom_diag/impl/atom_diag.cpp
krivenko: I changed the key type to `fock_state_t` and added the new constructor.

c++/triqs/atom_diag/impl/atom_diag.cpp
krivenko: Good catch, there should have been a `break` instead of the `continue`.
There are, however, two issues with the proposed solution.

- Checking for `std::abs(x) > 0.0` is redundant, `foreach()` calls its argument only for the non-vanishing amplitudes.
- I needed `found_nonzero_element` to make sure that zeroing of `m` is done only once. Otherwise the matrix elements could get wiped out at every `i_index` iteration.

c++/triqs/atom_diag/impl/atom_diag.cpp
```diff
@@ -107,17 +108,43 @@ namespace triqs {
...
+
+        // The initializer for i_index is needed here because of the Core Language Defect #2313.
+        // https://stackoverflow.com/a/46115028
+        foreach(final_st, [&, i_index = i_index, j_count = 0](fock_state_t j, scalar_t x) mutable {
+          assert(++j_count == 1);
+          for(auto const& sp : sub_hilbert_spaces) {
+            if(sp.has_state(j)) {
+              Bp = sp.get_index();
+
+              if(not found_nonzero_element) {
```
Wentzell: We could get rid of `found_nonzero_element` by checking if `m` is empty?
```suggestion
              if(m.empty()) {
```

Wentzell (commented):
Thank you @krivenko for updating the PR!
I had another read and only one final comment, see below.

c++/triqs/atom_diag/impl/atom_diag.cpp
Wentzell: I see, the foreach explicitly calls the `prune()`. C.f. comment below for the `found_nonzero_element`.

Wentzell:
Merging this, thank you again @krivenko !

PR #841, opened 2022-03-07
[stat] add global count to accumulator

phdum:
* rename log_bin_count to data_input_count as accessor
* add to test
* fix spelling errors in code / doc
* add nodiscard from clang-tidy
* clang-format


Wentzell:
Thank you @phdum for this API improvement and cleanup!

PR #842, opened 2022-03-24
Prototype for report measurements

hmenke:


Wentzell:
@hmenke Thank you for the PR! I added changes as we discussed.

Wentzell:
@parcollet Could you have a quick look at these changes?

hmenke:
Currently `mc.run` is private so there is no way to pass `do_measure = true` to the warmup.

c++/triqs/mc_tools/mc_measure_set.hpp
```diff
@@ -154,8 +155,8 @@ namespace triqs {
 
       std::string report() const {
         std::string result{};
-        for (auto &nmp : m_map) {
-          result += " " + nmp.first + "=" + nmp.second.report();
+        for (auto &[name, m] : m_map) {
+          if (m.report_measure) result += m.report() + "\n";
```
hmenke, line 165: Should the measure print its own name or should the registered name be used? In the latter case this change is necessary.
```suggestion
          if (m.report_measure) result += name + "=" + m.report() + "\n";
```

Wentzell: I would leave it up to the measurement how to format its report string and what to include. This gives us more flexibility.

Wentzell:
> Currently `mc.run` is private so there is no way to pass `do_measure = true` to the warmup.

What about calling instead `warmup` and/or `accumulate`?

hmenke:
`warmup` hardcodes `do_measure=false` and `accumulate` writes "Accumulating..." to stdout.

EDIT: See also our current integration with CT-INT https://github.com/hmenke/ctint/commit/a18720a4da9f5dbbd5ac288ac6a3667c4f3df793

Wentzell:
I have updated the PR to expose the `run` function in the public interface.

the-hampel (approved):

PR #843, opened 2022-03-24
Higher rank updates

hmenke:


Wentzell:
Thank you @hmenke and @marcel-klett, this was merged in a776efb6e and edc1668db

PR #845, opened 2022-05-27
use poll barrier in python to reduce CPU load per default as done in the C++ mpi layer

the-hampel:
same poll barrier implementation as in the C++ layer in triqs. Reduces CPU load on waiting MPI ranks drastically. Default wait time is 1 msec as in the C++ implementation. passing 0.0 or `None` to the barrier function call will use the default mpi.barier() call of mpi4py. I added a very brief test in one of the few mpi enabled python tests to check the function and cleaned up a few white spaces.

### DIFF
--- a/python/triqs/utility/mpi_mpi4py.py
+++ b/python/triqs/utility/mpi_mpi4py.py
@@ -51,7 +51,7 @@ def recv(source = 0):
 def barrier(poll_msec=1):
     """
     Use asynchronous synchronization, otherwise mpi.barrier uses up all the CPU time during
-    the run of subprocess. if poll_interval is 0.0 or None the default mpi barrier of mpi4py
+    the run of subprocess. if poll_msec is 0.0 or None the default mpi barrier of mpi4py
     is used.
     Parameters
     ----------
--- a/python/triqs/utility/mpi_mpi4py.py
+++ b/python/triqs/utility/mpi_mpi4py.py
@@ -55,7 +55,7 @@ def barrier(poll_msec=1):
     is used.
     Parameters
     ----------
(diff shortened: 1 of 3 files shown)


python/triqs/utility/mpi_mpi4py.py
```diff
@@ -47,33 +47,47 @@ def recv(source = 0):
     #print "node ", rank, " receives ", r, " from node ", source
     return r
 
-def barrier() : 
-    world.barrier()
+
+def barrier(poll_msec=1):
+    """
+    Use asynchronous synchronization, otherwise mpi.barrier uses up all the CPU time during
+    the run of subprocess. if poll_interval is 0.0 or None the default mpi barrier of mpi4py
```
Wentzell:
```suggestion
    the run of subprocess. if poll_msec is 0.0 or None the default mpi barrier of mpi4py
```

python/triqs/utility/mpi_mpi4py.py
```diff
@@ -47,33 +47,47 @@ def recv(source = 0):
...
+    is used.
+    Parameters
+    ----------
+    poll_interval: float, time step for pinging the status of the sleeping ranks in msec
```
Wentzell:
```suggestion
    poll_msec: float, time step for pinging the status of the sleeping ranks in msec
```

Wentzell (changes requested):
Thank you @the-hampel!

This looks good up to two minor issues in the doc strings.

the-hampel:
thanks! Yes I changed the name of the variable midway to make it more consistent with the C++ part and forgot to change the doc string...

PR #852, opened 2022-07-02
[doc] Slightly extend description of SOM on the applications page

krivenko:
SOM 2.0 comes with a new set of major features, which I would like to be mentioned on the 'Applications based on TRIQS' page.


Wentzell:
Thank you @krivenko for this update!

PR #855, opened 2022-07-12
Fix TB_from_pythTB function

the-hampel:
* create H_ij(-R) from [H_ji(+R)]* when loading pythTB object
* create stronger test for TB_from_pythTB function
* check hermiticity of hoppings when creating tight binding object


Wentzell (approved):
Thank you @the-hampel, this all looks good!

phibeck (approved):
Looks good, thank you!

PR #857, opened 2022-08-25
[cray-mpich] add support by EnvVar

HugoStrand:
Add environment variable in python MPI detection to support `cray-mpich`.

The equivalent change in C++ is done in TRIQS/mpi, see https://github.com/TRIQS/mpi/pull/10


Wentzell:
Thank you @HugoStrand for this contribution.

PR #864, opened 2022-11-17
Gf class matrix multiplication and right multiplication priority

GingrasO:
Dear all,

Following numpy's convention, the matrix multiplication operator should be @, while * performs elementwise multiplication. Thus I have added the set of operation corresponding to matrix multiplications to the Gf class. I simply changed the names from `mul` to `matmul` and created the `mul` functions such that they call the `matmul` functions. In other words, the Gf objects works the same, but now one can also perform the @ multiplication if used to numpy. However, it is not possible to do elementwise multiplication.

Moreover, I noticed the the `__rmul__` function of the Gf class was never called when the left operator was a numpy array. That is because the numpy array has a `__mul__` function that is instead called and leads to errors. To prioritize the Gf functions, I have added a `__array_priority__` value to the Gf class, which makes it call its `__rmul__` before numpy's `__mul__`.

### DIFF
--- a/python/triqs/gf/gf.py
+++ b/python/triqs/gf/gf.py
@@ -104,6 +104,7 @@ class Gf(metaclass=AddMethod):
     """
     
     _hdf5_data_scheme_ = 'Gf'
+    __array_priority__ = 10000 # Makes sure the operations of this class are applied as priority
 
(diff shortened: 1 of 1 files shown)


Wentzell:
Thank you @GingrasO for these improvements!

Issue #866, opened 2023-01-06
np.int is depracted (np v1.20) / removed (np v1.24)  but used in various tests


the-hampel:
In numpy version 1.24 `np.int` and `np.complex` have been removed as data type. They are deprecated since version 1.20. With numpy 1.24 and newer the following tests are failing:
```
 41 - Py_interface (Failed)
225 - py_gf_base_op (Failed)
```
will make a small bug fix soon.

Wentzell:
@the-hampel This can be closed now?

the-hampel:
yes! Has been solved with 200a4e17fdf9869c7c8551051ad6f738e3492927 and https://github.com/TRIQS/h5/commit/4a565c0fc59220a6603a90f4b1dc8ba8f7de9392 .

PR #867, opened 2023-01-09
[fix] np.int etc. removed in numpy 1.24

the-hampel:
np.int / np.float / np. complex are  deprecated (np v1.20) / removed (np v1.24) but used in various tests. Fixes issue #866 (https://github.com/TRIQS/triqs/issues/866)

Also fix some triqs internal deprecation warning regarding GfImFreq and GfImTime


Wentzell:
Looks good, thank you @the-hampel

PR #870, opened 2023-01-26
tests: cover the higher-rank updates implementation

hmenke:
Works on a776efb6e8e0d9312fc3a79184d3f684b690fa1d
Broken since edc1668db666a58b1c7e6a0b21fc2407c4b853ed


hmenke:
@Wentzell Maybe we can merge this just to get the tests up and running? For the case of regular double insertion `k` will always be `2` so there will not be any excessive `wk` resizing. The problem only occurs if updates with different `k` are interleaved (a situation which currently doesn't work at all).

Wentzell:
Thank you @hmenke for adding the tests, those were now merged into unstable, together with a proper fix of the issue.

PR #871, opened 2023-01-30
make res of trace_rho_op real if op + rho are hermitian

the-hampel:
* add function to check if a triqs many body operator is hermitian
* check if input density matrix and operator to trace_rho_op are hermitian and return real part of averaged value

### DIFF
--- a/c++/triqs/atom_diag/impl/functions.cpp
+++ b/c++/triqs/atom_diag/impl/functions.cpp
@@ -80,7 +80,7 @@ namespace triqs {
       auto is_block_matrix_hermitian = [](ATOM_DIAG_T::block_matrix_t const &bm) {
         return std::all_of(bm.begin(), bm.end(), [](ATOM_DIAG_T::matrix_t const &mat) { return max_element(abs(mat - dagger(mat))) == 0.0; });
       };
-      if (operators::is_op_hermitian(op) && is_block_matrix_hermitian(density_matrix)) {
+      if (is_op_hermitian(op) && is_block_matrix_hermitian(density_matrix)) {
         return std::real(result);
       } else
         return result;
--- a/c++/triqs/atom_diag/impl/functions.cpp
+++ b/c++/triqs/atom_diag/impl/functions.cpp
@@ -76,7 +76,14 @@ namespace triqs {
           if (b_m.first == sp) result += x.coef * trace(b_m.second * density_matrix[sp]);
         }
       }
-      return result;
+      // check if op && density matrix is hermitian then return real(result)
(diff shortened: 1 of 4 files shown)


the-hampel:
We could expose the function `is_op_hermitian` to python?

c++/triqs/atom_diag/impl/functions.cpp
```diff
@@ -76,7 +76,14 @@ namespace triqs {
           if (b_m.first == sp) result += x.coef * trace(b_m.second * density_matrix[sp]);
         }
       }
-      return result;
+      // check if op && density matrix is hermitian then return real(result)
```
Wentzell:
```suggestion
      // check if op && density matrix are hermitian, if so return real(result)
```

c++/triqs/atom_diag/impl/functions.cpp
```diff
@@ -76,7 +76,14 @@ namespace triqs {
...
+      auto is_block_matrix_hermitian = [](ATOM_DIAG_T::block_matrix_t const &bm) {
+        return std::all_of(bm.begin(), bm.end(), [](ATOM_DIAG_T::matrix_t const &mat) { return max_element(abs(mat - dagger(mat))) == 0.0; });
+      };
+      if (operators::is_op_hermitian(op) && is_block_matrix_hermitian(density_matrix)) {
```
Wentzell: Argument dependent lookup should trigger here
```suggestion
      if (is_op_hermitian(op) && is_block_matrix_hermitian(density_matrix)) {
```

Wentzell (changes requested):

c++/triqs/atom_diag/impl/functions.cpp
```diff
@@ -76,7 +76,14 @@ namespace triqs {
           if (b_m.first == sp) result += x.coef * trace(b_m.second * density_matrix[sp]);
         }
       }
-      return result;
+      // check if op && density matrix are hermitian, if so return real(result)
+      auto is_block_matrix_hermitian = [](ATOM_DIAG_T::block_matrix_t const &bm) {
+        return std::all_of(bm.begin(), bm.end(), [](ATOM_DIAG_T::matrix_t const &mat) { return max_element(abs(mat - dagger(mat))) == 0.0; });
+      };
+      if (is_op_hermitian(op) && is_block_matrix_hermitian(density_matrix)) {
+        return std::real(result);
```
Wentzell, line 91: Maybe you can add a small comment as to why this is even necessary?

the-hampel: this commit below

Wentzell (approved):

PR #872, opened 2023-02-02
fix: bug in is_convertible of block2_gf

the-hampel:
the is_convertible function did not return immediately if the obj was not a block2_gf. Not clear why this was written likes this, but this caused troubles since the set PyErrString was then not passed on properly. This resulted in segmentation faults when calling for example make_hermitian(1) instead of throwing an error.


Wentzell (approved):

Wentzell:
Thank you @the-hampel for this fix!

PR #873, opened 2023-03-14
[mpi-python] python/triqs.utility.mpi.all_reduce

jrbp:
Previous python all_reduce version did not use first or third argument at all. Now comm and op are optional arguments, actually work, and the signature better matches the c++ api.


Wentzell (approved):
Thank you @jrbp for this contribution!

I will merge as soon as CI has passed.

PR #874, opened 2023-04-19
plot: fix histogram plotting via oplot

the-hampel:
- functionality to add python only function to a cpp2py generated class does not exist anymore
- add global dict / table to store plotting functions for C++ classes
- register histogram plot function in init


Wentzell:
Thank you @the-hampel for restoring this functionality!

Issue #881, opened 2023-06-02
No mesh compatibility check on β when assigning Green's function defined by blocks


Crampon16:
### Description

When two Green's functions are defined by block (e.g. "up", "dn") and have different temperatures, assigning the content of one to the other does not raise an exception.

### Steps to Reproduce

```python
from triqs.gf import *

g1 = GfImFreq(beta=10, indices=[0])
g2 = GfImFreq(beta=20, indices=[0])

G1 = BlockGf(block_list=[g1,g1], name_list=['up','dn'], make_copies=True)
G2 = BlockGf(block_list=[g2,g2], name_list=['up','dn'], make_copies=True)

G1['up'] << G2['dn']  # complains
G1 << G2              # does not complain !!
```

**Expected behavior:** an Assertion Error because of mesh compatibility is raised

**Actual behavior:** ``G2``'s data is copied to ``G1``, (without altering the original mesh's temperature).

### Versions
```
You are using the TRIQS library version 3.0.0


You are using the git hash 
```
I am using MacOS Mojave v 10.14.6, and have also been able to reproduce on various Ubuntus.

### Additional Information

Jupyter notebook to quickly reproduce the error:
[bug_no_beta_check.ipynb.zip](https://github.com/TRIQS/triqs/files/11626148/bug_no_beta_check.ipynb.zip)

the-hampel:
Dear @Crampon16,
this is indeed a problem I would say. I don't see why this would be the desired behavior to overwrite the block gf with a different mesh. The problem is also existent in TRIQS 3.1.x and on the current unstable branch.

Wentzell:
Fixed by d64d76754

PR #883, opened 2023-06-05
[mpi] add envar check to force MPI init

the-hampel:
Add check for environment variable `TRIQS_FORCE_MPI_INIT` to allow enforcing the MPI Init for specific environments and test cases. Should address TRIQS/cthyb issue #166 (https://github.com/TRIQS/cthyb/issues/166)

PR #884, opened 2023-06-05
fix issue 881

the-hampel:
merge after #883. 

Fixes issue #881 by adding asserts to all block gf types similar to the normal gf to ensure mesh compatibility between each block that is assigned.


Wentzell (approved):

PR #890, opened 2023-06-20
[gf] add array_priority also to BlockGf / Block2Gf

the-hampel:
multiplication of type numpy array * BlockGf was broken because ladd operation of numpy was priorized over the triqs implementation. Adding the numpy array_priority attribute solves this. Compare with Gf class implementation.


Wentzell (approved):
Thank you @the-hampel  for this improvement!

Wentzell:
Merged in 0bb7b1489

PR #892, opened 2023-06-28
[gf] fix density function for MeshReFreq Gf

the-hampel:
density function always returned 0.0 as result of recent changes. 

Additionally this commit strengthens the test for density(), density(beta), and total_density. Added tests in C++ and Python.


Wentzell (approved):

Wentzell:
Thank you @the-hampel for this fix!

PR #896, opened 2023-07-18
Added the sympyfy function alongside its necessary dependencies

Collins-kariuk:
The sympyfy function outputs a reduced analytical or numerical Hamiltonian in momentum space utilizing the Fourier series. The input of this function is a TRIQS tight-binding model that one can get from converting their model from either a Wannier90 model or PythTB model using TRIQS's handy function that enables said conversion.

This function has been tested on 2 compounds with different symmetries. These compounds include a Strontium compound (SrCaCu2O4) and a Palladium compound (PdCoO2). This function mainly follows from the equation in the attached screenshot. This function will help anyone, especially experimentalists, better understand their results in the tight-binding model landscape by visualizing a reduced analytical expression (and a numerical expression should they wish).
Tight-Binding Hamiltonian in Momentum-Space from Real Space using the Fourier Series

### DIFF
--- a/python/triqs/lattice/utils.py
+++ b/python/triqs/lattice/utils.py
@@ -275,9 +275,12 @@ def sympyfy(tb_lat_obj, analytical = True):
         numerical form or reduced analytical form depending on what the user chooses. The default output, 
         which depends on the optional analytical flag or parameter, is the reduced analytical form
 
+        #TODO: Explain what the numerical form means. Only depends on k and no other variables? 
+        # TODO: versus the analytical form. 
+
+
     """
     import sympy as sp
-    from sympy import exp
 
     # imaginary number
     I = sp.I
@@ -370,7 +373,7 @@ def sympyfy(tb_lat_obj, analytical = True):
         """
         for sublist in matrix:
             for element in sublist:
-                if element.is_complex and element.has(exp):
+                if element.is_complex and element.has(sp.exp):
                     return True
         return False
     
--- a/python/triqs/lattice/utils.py
+++ b/python/triqs/lattice/utils.py
@@ -246,3 +246,161 @@ def TB_from_pythTB(ptb):
                     orbital_names=[str(i) for i in range(ptb.get_num_orbitals())])
 
     return TBL
+
+# importing necessary dependencies
+from itertools import product as itp
+from pythtb import *
+# from triqs.lattice.tight_binding import TBLattice # TRIQS needs to be installed for this dependency to work
+import sympy as sp
+from sympy import *
+import warnings
+
+def sympyfy(tb_lat_obj, analytical = True):
+    r"""
+    returns the analytical form of the momentum space hamiltonian of the tight-binding model,  
+    from a tight-binding lattice object, by utilizing Fourier series
+    
+    Parameters
+    ----------
+    tb_lat_obj: triqs TBLattice object
+        triqs tight binding object
+    analytical: boolean, default = True
+        a boolean which will cause the function will return an analytical Hamiltonian, when true, and 
+        an numerical Hamiltonian otherwise
+    
+    Returns
+    -------
+    Hk_numerical: NumPy Array
+        the hamiltonian of the tight-binding model in momentum space in numerical form
+    Hk: NumPy Array
+        the hamiltonian of the tight-binding model in momentum space in reduced analytical form
+
+    """
+
+    # imaginary number
+    I = sp.I
+
+    # matrix from the axis directions in momentum space
+    kx, ky, kz = sp.symbols("kx ky kz", real = True)
+    k_space_matrix = sp.Matrix([kx, ky, kz])
+
+    # symbolic dot product representation between the lattice unit vectors
+    # and the momentum space matrix
+    a1k, a2k, a3k = sp.symbols("a1k a2k a3k", real = True)
+    lattice = sp.Matrix([a1k, a2k, a3k])
+
+    # the number of orbitals involved in the electron hoppings
+    num_orb = tb_lat_obj.n_orbitals
+
+    # dictionary containing details about the hopping of the electrons
+    TB_lat_obj_hops = tb_lat_obj.hoppings 
+
+    # maximum hopping distances of electrons in each of the axial directions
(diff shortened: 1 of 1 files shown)


python/triqs/lattice/utils.py
```diff
@@ -246,3 +246,161 @@ def TB_from_pythTB(ptb):
...
+
+    # dealing with the numerical Hamiltonian
+    # we convert it to a SymPy matrix to use the substitutions method available in SymPy
+    Hk_numerical = sp.Matrix(Hk)
+    TB_lat_obj_units = tb_lat_obj.units
+    TB_lat_obj_units_transpose = np.transpose(TB_lat_obj_units)
+    
+    # obtaining unit vectors
+    # reduce floating point precision to 3 decimal places
+    a1 = np.around(TB_lat_obj_units_transpose[0], decimals = 3)
```
the-hampel, line 346: does it make sense to let the user define the precision at this point?

python/triqs/lattice/utils.py
```diff
@@ -246,3 +246,161 @@ def TB_from_pythTB(ptb):
...
+    ----------
+    tb_lat_obj: triqs TBLattice object
+        triqs tight binding object
+    analytical: boolean, default = True
+        a boolean which will cause the function will return an analytical Hamiltonian, when true, and 
+        an numerical Hamiltonian otherwise
+    
+    Returns
+    -------
+    Hk_numerical: NumPy Array
```
the-hampel: This looks as if the function gives back to values, but it only gives back one value. Just list in Return one object and explain in the text that the object given back depends on the `analytical` flag.

python/triqs/lattice/utils.py
```diff
@@ -246,3 +246,161 @@ def TB_from_pythTB(ptb):
...
+
+# importing necessary dependencies
+from itertools import product as itp
+from pythtb import *
+# from triqs.lattice.tight_binding import TBLattice # TRIQS needs to be installed for this dependency to work
+import sympy as sp
+from sympy import *
+import warnings
+
+def sympyfy(tb_lat_obj, analytical = True):
```
the-hampel, line 258: I am not sure what is the meaning of the numerical form? Just the Fourier transform of H(R)? On what kind of mesh?

python/triqs/lattice/utils.py
```diff
@@ -246,3 +246,161 @@ def TB_from_pythTB(ptb):
                     orbital_names=[str(i) for i in range(ptb.get_num_orbitals())])
 
     return TBL
+
+# importing necessary dependencies
+from itertools import product as itp
+from pythtb import *
+# from triqs.lattice.tight_binding import TBLattice # TRIQS needs to be installed for this dependency to work
+import sympy as sp
+from sympy import *
```
the-hampel: It would be good to include import module statements needed for the function in the function itself in the beginning, and more general things like itertools, and warnings at the top of the `uitls.py` file.

python/triqs/lattice/utils.py
```diff
@@ -246,3 +246,161 @@ def TB_from_pythTB(ptb):
...
+    
+    # performing the numerical dot product substitutions
+    Hk_numerical = Hk_numerical.subs(a1k, a1k_numerical)
+    Hk_numerical = Hk_numerical.subs(a2k, a2k_numerical)
+    Hk_numerical = Hk_numerical.subs(a3k, a3k_numerical)
+
+    # converting the numerical Hamiltonian to a NumPy array from a SymPy matrix
+    Hk_numerical = np.array(Hk_numerical)
+
+    def has_complex_exponential(matrix):
```
the-hampel: maybe clarify that the functions are only for sympy by adding sympy to the function name. Consider making the functions iternal only by adding `_` in front.

python/triqs/lattice/utils.py
```diff
@@ -246,3 +246,161 @@ def TB_from_pythTB(ptb):
...
+    # basis of the 5D tensor real-space Hamiltonian
+    Hrij = np.zeros((num_cells_x, num_cells_y, num_cells_z, num_orb, num_orb), dtype = sp.exp)
+
+    # looping through the hopping parameters of the electrons involved in the inter-orbital hoppings
+    # key represents the cell coordinates of where the electrons hop to relative to the home unit cell
+    # hopping represents the matrix with the embedded hopping amplitudes
+    for key, hopping in TB_lat_obj_hops.items():
+        rx, ry, rz = key
+        # reduce floating point precision of hopping parameters to 3 decimal places
+        hopping = np.around(hopping, decimals = 3)
```
the-hampel, line 318: Does it make sense to let the user specify the precision?

the-hampel (changes requested):
Hi @Collins-kariuk ,
thank you for the PR. Please consider my comments below. Also could you please create the PR against the triqs/unstable branch instead. Thank you.

python/triqs/lattice/utils.py
Collins-kariuk: The default according to Wannier is 10^-6 but the user can change it, there will be an optional parameter to change the precision.

python/triqs/lattice/utils.py
Collins-kariuk: Fixed.

python/triqs/lattice/utils.py
Collins-kariuk: It is the Fourier transform of H(R) which only depends on **k**. There is no mesh because **k** is not specified. 
In some sense, it is also analytical because it depends on trigonometric functions but there are no lattice parameters or displacement vectors.

python/triqs/lattice/utils.py
Collins-kariuk: Fixed.

python/triqs/lattice/utils.py
Collins-kariuk: Fixed.

Collins-kariuk:
Olivier, Sophie, and I thought it best if I create a new branch, make the necessary changes to utils.py within that branch, and make a pull request from my branch to the upstream unstable branch. So I am going to close this pull request.

PR #899, opened 2023-07-23
Update requirements.

RemiLacroix-IDRIS:
Add GMP that was not listed previously and update Boost minimum version to match the CMake scripts.


Wentzell:
Thank you @RemiLacroix-IDRIS for this correction!

RemiLacroix-IDRIS:
You're welcome!

Issue #679, opened 2019-02-04
Feature request [MC]: -1 param for n_cycles


amoutenet:
Hi all,

In some cases I'd like my MC to run for a given time regardless of the number of cycles it needs. To achieve this, I set a max_time for the Monte Carlo, and a ridiculous amount of cycles.
However, this is error prone as you might actually reach that number of cycles without noticing it and have a run time shorter than the one you were actually targeting.

Therefore, would it be possible to implement n_cycles = -1? This could work the same way as max_time = -1 does, meaning running max_time regardless of the actual number of cycles.
Of course this would require to throw an error if a user tries to put n_cycles = -1 and max_time = -1.

Cheers,
Alice

Wentzell:
Dear Alice,

Thank you for the suggestion! We agree that this would make the API more consistent.
@parcollet The steps to be taken are

- Type of `n_cycles` in mc_generic should be `long` not `uint64_t`
- Check for `n_cycles == -1` in the loop

Wentzell:
Fixed in https://github.com/TRIQS/triqs/issues/679

PR #902, opened 2023-08-08
Allow negative n_cycles in mc_generic

Wentzell:
- Allow negative values for n_cycles in mc_generic: Run until call_back triggers or signal is send
- Report only cycles and no ETA
- Change unsigned integer types to signed ones in mc_generic implementation

C.f. #679


the-hampel (commented):
These changes look good to me. Small questions before merging:  

- Are these covered by any test? If not, should we write a test, i.e. test negative n_cycles and stop via a callback? 
- should it be more clear that any negative number results in the same outcome?

the-hampel (approved):

PR #904, opened 2023-08-23
Partially revert c5b8670, restore k_space_path TPRF compat

Wentzell:
Suggestion by @HugoStrand, to be discussed.

I think the API is less clear this way, ticks should contain the actual tick positions along the way and not the edges?

### DIFF
--- a/python/triqs/lattice/utils.py
+++ b/python/triqs/lattice/utils.py
@@ -24,12 +24,12 @@ import warnings
 __all__ = ['k_space_path', 'TB_from_wannier90']
 
 
-def k_space_path(paths, num=101, bz=None, relative_coordinates=True, return_ticks=False):
+def k_space_path(segments, num=101, bz=None, relative_coordinates=True):
     """ Generate an array of k-vectors along a path defined by a list of pairs of k-vectors
 
     Parameters
     ----------
-    paths : list of pairs of three-vectors of floats
+    segments : list of pairs of three-vectors of floats
        List of pairs of k-vectors in reciprocal units to create a path in-between.
     num : int, default=100
        Number of k-vectors along each segment of the overall path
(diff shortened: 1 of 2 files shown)


HugoStrand:
I disagree with the statement. I would like the tick positions returned to be all high symmetry points in the plot, so that the returned vector can be passed directly to a `plt.ticks(...)` call, see below.

Please note that this is a new and optional feature that I have added, so we do not yet have TRIQS users using this.

Including the first and last point will enable cleaner examples in the documentation by removing the `np.concatenate` clutter we currently have here for example:

https://triqs.github.io/tprf/unstable/user_guide/Square%20lattice%20susceptibility.html#Dispersion-\epsilon(\mathbf{k})-and-Fermi-surface

With the proposed change the code is simplified to
```python
G = [0.0, 0.0, 0.0]
X = [0.5, 0.0, 0.0]
M = [0.5, 0.5, 0.0]

path = [(G, X), (X, M), (M, G)]

from triqs.lattice.utils import k_space_path

k_vecs, k_plot, k_ticks = k_space_path(
    path, num=32, bz=H_r.bz, 
    relative_coordinates=False, return_ticks=True)

e_k_interp = np.vectorize(lambda k : e_k(k).real, signature='(n)->()')

plt.plot(k_plot, e_k_interp(k_vecs))
plt.xticks(k_ticks, labels=[r'$\Gamma$', '$X$', '$M$', r'$\Gamma$']) # Here can pass k_ticks directly without additional operations.
plt.ylabel(r'$\epsilon(\mathbf{k})$')
plt.grid(True)
```

This produces the standard kind of high-symmetry path figure as in the example above.

I want this for the TRIQS tutorials next week.

While I wait for this pull request I will add a compatibility layer in TPRF so that I can finish the tutorials.

HugoStrand:
Here is the promised TPRF compatibility layer, 
https://github.com/TRIQS/tprf/commit/2a2010df68d6cddf76cd7e16595cc082bf0e0297

the-hampel (approved):
I think also giving back end points makes sense to me. Usually I obtained the tick distances via: 
```
np.concatenate(([0],k[num-1::num]))
```
so the new functionality makes most sense to me including end points. 

Thank you for the addition.

Wentzell:
Thank you @HugoStrand @the-hampel for clarifying. I agree with the points you made.

In that case the calculation of the ticks is more involved than just doing `dist[num:num]`, which was the initial reason I decided not to return them. I iterated to commit and changed it to always return the ticks.

the-hampel (commented):
Looks good to me. Are we all fine with the breaking change of returning three objects instead of two? I think for me it should be only small changes to codes.

Issue #918, opened 2023-09-22
AtomDiag passes complex operator to AtomDiagReal if numerically real


thenoursehorse:
**Expected behavior:** If the Hamiltonian passes the check that it is real, it should pass a real operator to AtomDiagReal.

**Actual behavior:** The Hamiltonian passes the check that it is real, and passes a complex operator to AtomDiagReal.

Sometimes I construct operators from a complex Hamiltonian that has zero imaginary values at the start of a self-consistent loop, but will become nonzero later in the cycle. It is not hard to check myself and pass the correct object, but fixing it may be beneficial others.

Below is code that reproduces the error

```python
from triqs.atom_diag import AtomDiag
from triqs.operators import n

h = (1 + 0*1j) * n('up', 0) * n('dn', 0)
fops = [('up', 0), ('dn', 0)]
AtomDiag(h, fops)
```

### Version
TRIQS library version 3.2.0
git hash 90c8f8c257be434bc4560f4bbc518905c4d226ea

Wentzell:
Fixed by https://github.com/TRIQS/triqs/pull/919

PR #919, opened 2023-09-22
fix: python AtomDiag force real if real

thenoursehorse:
Fixes triqs/triqs#918

Forces h to be real when passed to AtomDiagReal

The unit test might also be useful to check that python is passing the correct objects to the c++ constructors.


Wentzell:
Thank you @thenoursehorse for this improvement. I have iterated on the PR and pushed a number of changes. Please have a look and let me know if you agree.

Wentzell (approved):

thenoursehorse:
It looks good to me. Thanks @Wentzell.

Issue #920, opened 2023-09-22
Update cmake version check to work with ranges


thenoursehorse:
**Expected behavior:** When using cmake's `find_package(TRIQS "min_version...max_version")` and a found TRIQS version `>max_version`, TRIQS should tell cmake that there is no suitable version found.

**Actual behavior:** TRIQS tells cmake a suitable version is found when there is not, and will try build against that version.

This issue was discussed here: krivenko/ezARPACK#7.

From the [cmake manual](https://cmake.org/cmake/help/latest/command/find_package.html#config-mode-version-selection) it specifies that when a range of versions is given, it essentially sets `PACKAGE_FIND_VERSION` to the minimum version in the range for backwards compatability. The TRIQS version checking only uses this variable, instead of the `PACKAGE_FIND_VERSION_MIN` and `PACKAGE_FIND_VERSION_MAX` variables.

The `TRIQSConfig-version.cmake.in` file needs to include the below to work with ranges:

```cmake
if (PACKAGE_FIND_VERSION_MAX VERSION_LESS PACKAGE_VERSION)
 set(PACKAGE_VERSION_UNSUITABLE TRUE)
endif()

if (PACKAGE_FIND_VERSION_MIN VERSION_GREATER PACKAGE_VERSION)
 set(PACKAGE_VERSION_UNSUITABLE TRUE)
endif()
```

Or, at least the first one, because the `min_version` is handled by the `PACKAGE_FIND_VERSION` variable. But there is likely no harm in using both checks.

### Additional Information
TRIQS minimum cmake version is `3.20`. Ranges in `find_package` was introduced in cmake version `3.19`. 

### Version
TRIQS library version 3.2.0
git hash 90c8f8c257be434bc4560f4bbc518905c4d226ea

Wentzell:
Fixed by https://github.com/TRIQS/triqs/pull/921

PR #921, opened 2023-09-22
fix: range checking for cmake find_package

thenoursehorse:
Fixes triqs/triqs#920.

I have not included a test to show that the version checking works. I can write up a mock c++ project and attach if necessary.


Wentzell (approved):
Thank you @thenoursehorse for this fix!

PR #924, opened 2023-10-28
Enable compiler warnings

Thoemi09:


Wentzell (approved):
Thank you @Thoemi09 for this big improvement!
Having all major compiler warnings enabled is an important step.

PR #935, opened 2024-03-06
[doc] Update build_triqs.sh to work for zsh on macos

Thoemi09:


the-hampel (approved):
I changed slightly the folder name of the clone repos for consistency. Otherwise this works still nicely in `bash`.

Wentzell (approved):

PR #936, opened 2024-04-04
[Gf] add constrained residual minimization dyson solver

the-hampel:
* add l2 tau norm function from cppdlr
* add tests for l2 tau norm
* fix loss of prec warning in DLR tests
* add Python bindings for constructors of DLR meshes from other DLR meshes
* add constrained residual minimization (CRM) function to solve Dyson's equation https://arxiv.org/abs/2310.01266
* add tests for CRM Dyson solver

Example of the method for a 3orbital Bethe lattice test at n=0.8 filling with realistic QMC data at beta=1000 eV for D=1 (half-bandwidth):
crm_test_b1000


Wentzell (approved):
Thank you @the-hampel for this new feature!
This all looks good to me now, and you can go ahead with the merge.

PR #950, opened 2024-07-15
[feat] selectively add SF / PH terms in Kanamori

the-hampel:
This PR allows 
* to select specifically spin-flip and / or pair-hopping terms are added in the construction of the `h_int_kanamori` many-body operator. Additionally I run auto format on the python file. 
* fixes `x_data_view`: for MeshReFreq objects it would only give back a np array of mesh points for `X`
* extend `x_data_view` to MeshImTime
* speed up `x_data_view` by removing for loop to extract `X` values

### DIFF
--- a/python/triqs/gf/gf.py
+++ b/python/triqs/gf/gf.py
@@ -18,7 +18,7 @@
 # Authors: Michel Ferrero, Olivier Gingras, Alexander Hampel, Jonathan Karp, Manuel, Olivier Parcollet, Markus Richter, Hugo U. R. Strand, Nils Wentzell
 
 import itertools, warnings, numbers
-from functools import reduce # Valid in Python 2.6+, required in Python 3
+from functools import reduce  # Valid in Python 2.6+, required in Python 3
 import operator
 import numpy as np
 from . import mesh_product
@@ -27,7 +27,7 @@ from . import descriptors, descriptor_base
 from .mesh_product import MeshProduct
 from triqs.plot.protocol import clip_array
 from . import meshes
-from . import plot 
+from . import plot
 from . import gf_fnt, wrapped_aux
 from .mesh_point import MeshPoint
 from .matsubara_freq import MatsubaraFreq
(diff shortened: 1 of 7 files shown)


Wentzell:
Thank you @the-hampel for the changes.

Can you please separate the code changes from the formatting changes in the commit history?

the-hampel:
The commits are now separate into three parts: 1) formatting, 2) feature in hamiltonians.py, and 3) fix for `x_data_view` .

python/triqs/gf/gf.py
```diff
@@ -806,7 +806,14 @@ def x_data_view(self, x_window=None, flatten_y=False):
             If flatten_y is True and dim is (1, 1, *) it returns a 1d numpy array.
         """
 
-        X = [x.imag for x in self.mesh] if isinstance(self.mesh, meshes.MeshImFreq) else [x for x in self.mesh]
+        if isinstance(self.mesh, meshes.MeshReFreq):
+            X = np.linspace(self.mesh.w_min, self.mesh.w_max, len(self.mesh))
```
Wentzell: I don't think it is guaranteed that this generates the same floating point numbers as `[x.real for x in self.mesh]` ?

python/triqs/gf/gf.py
```diff
@@ -806,7 +806,14 @@ def x_data_view(self, x_window=None, flatten_y=False):
...
+        elif isinstance(self.mesh, meshes.MeshImFreq):
+            X = np.linspace(self.mesh(self.mesh.first_index()).imag, self.mesh(self.mesh.last_index()).imag, len(self.mesh))
```
Wentzell: Same as above

python/triqs/gf/gf.py
```diff
@@ -806,7 +806,14 @@ def x_data_view(self, x_window=None, flatten_y=False):
...
+        elif isinstance(self.mesh, meshes.MeshImTime):
+            X = np.linspace(0, self.mesh.beta, len(self.mesh))
```
Wentzell: Same as above

Wentzell (changes requested):
Dear @the-hampel

Thank you for these improvements! See comments below.

python/triqs/gf/gf.py
the-hampel: hm, yes that is probably true. Are you concerned about exact equality of the values? Is it necessary? But okay then we need to loop again right? I hoped to make this method also a bit faster by avoiding for loops.

Wentzell: I have moved the generation of the values array to C++, please have a look at the changes.

the-hampel (commented):
Thank you @Wentzell This looks good to me.

PR #961, opened 2024-08-27
[hilbert_space] Simplify algorithm of `space_partition::merge_subspaces()`

krivenko:
The current implementation of `space_partition::merge_subspaces()` uses recursion, `std::function` and repeatedly deletes elements from a `std::multimap`, which results in a significat runtime overhead.

More concerningly, I experienced a stack oveflow caused by the recursion while trying to run the autopartition algorithm for a bigger system.

The new algorithm simply iterates over subspace connections $i\mapsto f$ generated by a single operator $c^\dagger/c$, and merges all subspaces $f$ corresponding to the same $i$.

### DIFF
--- a/c++/triqs/hilbert_space/space_partition.hpp
+++ b/c++/triqs/hilbert_space/space_partition.hpp
@@ -146,42 +146,26 @@ namespace triqs {
           tmp_state(i) = amplitude_t(0.);
         }
 
-        // 'Zigzag' traversal algorithm
-        while (!Cd_connections.empty()) {
+        // Merge all 'out' subspaces corresponding to the same 'in' subspace
+        // in 'conn'.
+        auto merge_conn_targets = [this](auto const& conn) {
+          if(conn.empty()) return;
 
-          // Take one C^+ - connection
-          // C^+|lower_subspace> = |upper_subspace>
-          idx_t lower_subspace, upper_subspace;
-          std::tie(lower_subspace, upper_subspace) = *std::begin(Cd_connections);
-
-          // - Reveals all subspaces reachable from lower_subspace by application of
-          //   a 'zigzag' product C^+ C C^+ C C^+ ... of any length.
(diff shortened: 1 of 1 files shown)


Wentzell:
Dear @krivenko 
Thank you for proposing this simplification.

It is not immediately obvious to me why the new implementation will always lead to the same result as before.

In the previous case you were using the zigzag traversal starting with the upwards connection `i0->f0`.
At each step of the zig zag you were linking all final subspaces with the same initial subspace to the original `i0` or `f0` respectively. This would also link subspaces further down the zig zag line to the initial `i0` or `f0`.

This no longer applies in the new implementation, where all you are looking at is which final subspaces can be directly reached from the same initial one.

Can you please clarify? Happy to talk over zoom if thats easier.

krivenko:
> In the previous case you were using the zigzag traversal starting with the upwards connection i0->f0.
> At each step of the zig zag you were linking all final subspaces with the same initial subspace to the original i0 or f0 respectively. This would also link subspaces further down the zig zag line to the initial i0 or f0.
>
> This no longer applies in the new implementation, where all you are looking at is which final subspaces can be directly reached from the same initial one.

The ultimate goal of the algorithm is to make sure that each $c_i$ and $c^\dagger_i$ generates only one-to-one connections between subspaces. This condition can be checked and enforced independently for each operator -- there is, in fact, no real need to work with pairs $(c_i, c^\dagger_i)$ and to do any sort of back-and-forth iteration. Once the condition is fulfilled for an operator, it cannot be violated by any further subspace mergers.

Therefore, one can simply take an operator, merge all final subspaces corresponding to the same initial subspace, and proceed to the next operator.

krivenko:
@Wentzell I have fixed the algorithm following the conclusions of our recent discussion.

krivenko:
@Wentzell This PR seems to be ready, can we have it merged?

PR #963, opened 2024-09-12
Apply mpi::monitor changes to mc_generic

Thoemi09:


Wentzell:
Thank you @Thoemi09 for these compatibility changes. This makes the necessary API changes for compatibility against https://github.com/TRIQS/mpi/pull/20

PR #966, opened 2024-09-24
Enabled callback function in the accumulator

dominikkiese:
Implemented a callback in the (linear) accumulator, which is called after each recompression step if defined. It defaults to the calculation of autocorrelation times from the variances of the original and binned data. Based on PR#79 in TRIQS/nda (https://github.com/TRIQS/nda/pull/79).

### DIFF
--- a/c++/triqs/stat/accumulator.hpp
+++ b/c++/triqs/stat/accumulator.hpp
@@ -43,6 +43,10 @@ namespace triqs::stat {
       std::vector<T> bins;     // Bins with accumulated data (stores means)
       long count = 0;          // Total number of elements added to accumulator: Information only
 
+      // Callback function after compressing bins
+      using callback_t    = std::function<void(std::vector<T> const &)>;
+      callback_t callback = {};
+
       // static std::string hdf5_format() { return "linear_bins"; }
 
       lin_binning() = default;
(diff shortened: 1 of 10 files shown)


Wentzell:
Please reopen the PR using the https://github.com/TRIQS/triqs/tree/ACCUM branch, where I have rebased your changes onto unstable

c++/triqs/stat/accumulator.hpp
```diff
@@ -344,9 +357,44 @@ namespace triqs::stat {
...
+          // variance of the binned data
+          T delta                = bins[0] - mean_with_binning;
+          var_t var_with_binning = make_real(nda::hadamard(nda::conj(delta), delta));
+
+          for (auto i : range(1, bins.size())) {
+            delta = bins[i] - mean_with_binning;
+            var_with_binning += make_real(nda::hadamard(nda::conj(delta), delta));
+          }
+
+          var_with_binning /= bins.size();
```
Wentzell, line 388: Please use the `mean_and_err` function from the `mean_and_err.hpp` header here.
Adjust it accordingly to use `hadamard`.

dominikkiese: Reopened as #967. I would propose to not use the `mean_and_err` function since we want the variance of the sample not the error on the mean in this case (we thus do not need to assume that `sqrt(T)` is defined).

PR #967, opened 2024-09-26
Accum

dominikkiese:
Reopening of #966


Wentzell:
Thank you @dominikkiese for this improvement! I pushed some simplifications in my commit ffb12cc

Can you please look into creating a test of the callback functionality?

dominikkiese:
Commit looks good to me, using `nda::map` with the `abs_square` function makes the code much more compact! I will think about  a proper test.

PR #968, opened 2024-09-26
Update MC tools

Thoemi09:
* Clean up all files in`mc_tools`
* Modernize the type erasure implementation
* Split header into header + source and instantiate templates explicitly
* Introduce a struct for run parameters in `mc_generic`
* Make the run method more flexible by introducing new parameters, e.g. to run a minimum number of cycles, how often we check for exceptions, etc.
* Improve the information that is printed during a MC simulation


Wentzell (approved):
Thank you @Thoemi09 for this big improvement. This was now merged after rebase.

PR #973, opened 2024-11-24
Make GF's MPI routines non-lazy

Thoemi09:


Wentzell (approved):
Thank you @Thoemi09 for this big improvement and for incorporating all suggested changes. This looks great now!

PR #978, opened 2025-01-21
Python script for (offline) installation

Thoemi09:
Right now, the workflow is something like this:

- clone and archive on local machine: `python clone_archive_and_install.py -c -a`
- copy the archive to the remote machine
- extract: `tar xzf triqs_repositories`
- build and install: `python triqs_repositories/clone_archive_and_install.py -i -j 12`

The script takes a few command line arguments with which the git clone and cmake commands can be adjusted. Additionally, one can also change the `repos` dict directly in the python script.


Wentzell (approved):
Thank you @Thoemi09 for this improvement!

PR #979, opened 2025-01-31
MagInt description included

leopo:
I added a short description of MagInt in the ab initio tools section


Wentzell:
Thank you @leopo for this addition!

PR #982, opened 2025-03-11
[doc] Add edipack2triqs to the list of TRIQS-based applications

krivenko:


Wentzell (approved):
Thank you @krivenko for adding this.

PR #1001, opened 2026-03-25
Add c2py converters

Thoemi09:
- make them independent of cpp2py


Wentzell (approved):
Thank you @Thoemi09 for this improvement

PR #1003, opened 2026-04-03
Add direct evaluation of imaginary time Green's function with semicircular spectral function

jasonkaye:
With this pull request, syntax like
```
g_tau = Gf(mesh=MeshDLRImTime(beta, 'Fermion', w_max, eps), target_shape=[])
g_tau << SemiCircular(D)
```
now works correctly. Formulas/algorithms to compute the Green's function with a semicircular spectral function have been factored into a new file, and the descriptors.py file has been updated to use these. The imaginary time Green's function is computed from the Lehmann representation using a custom adaptive Gauss quadrature, which is tested for near-machine-precision agreement against automatic adaptive quadrature in scipy. The notation above is also tested by comparing against the analytic Matsubara frequency formula after Fourier transform.

### DIFF
--- a/python/triqs/gf/descriptors.py
+++ b/python/triqs/gf/descriptors.py
@@ -89,21 +89,16 @@ semicircle
         mu = self.chem_potential
         Id = 1. if len(G.target_shape) == 0 else numpy.identity(G.target_shape[0])
         if type(G.mesh) in [MeshImFreq, MeshDLRImFreq]:
-            def f(om_):
-                return g_semicirc_iw(om_ + mu, D) * Id
+            vals = g_semicirc_iw(G.mesh.values() + mu, D)
         elif type(G.mesh) in [MeshReFreq, MeshReFreqPts, MeshReFreqLog]:
-            def f(om_):
-                return g_semicirc_w(om_.real + mu, D) * Id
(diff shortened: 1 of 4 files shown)


Wentzell:
Thanks for the PR, @jasonkaye! I've pushed a few follow-up commits on top of this branch (`jasonkaye/triqs:unstable`, four commits starting at fc95b9db). The changes:

- fc95b9db — simplify the imaginary-time branch of `SemiCircular.__call__` (hoist `Id`, use `G.mesh.values()`, drop the per-point loop)
- 3499c3e4 — fix a pre-existing bug where `SemiCircular` on a matrix-valued Gf over a real-frequency mesh filled both diagonal and off-diagonal entries instead of `g(ω) · I`; added a test covering all three supported meshes
- 133dcf81 — unify the three mesh-type branches to share a single vectorized `G.data` assignment
- abbc38e9 — minor clarity fix in `g_semicirc_iw` (name the real frequency instead of `(iw/1j)**2`)

All `py_gf_*` tests pass locally. Would you mind taking a look when you get a chance?

Wentzell (approved):

jasonkaye:
Looks good to me, thanks.

PR #1005, opened 2026-04-14
Port to c2py + clair

Thoemi09:
This is a draft PR for the port to c2py + clair. So far, everything has been ported and the tests are passing.

### Open questions/todo:

- How to split the GF python modules into smaller submodules?
- Python documentation needs to be generated and updated. This goes hand in hand with the doxygen documentation in C++.
- What kind of cmake options should we use? They should be consistent with other core libraries and applications.

### Dependencies

As of now, we need the following commits/branches for a successful build (they will be merged into their respective unstable branches): 
- https://github.com/TRIQS/nda/pull/127
- https://github.com/flatironinstitute/clair/pull/23 (if clair uses llvm-22)
- https://github.com/flatironinstitute/c2py/pull/19 (if clang < 21 is used)


Wentzell (approved):
Thank you @Thoemi09 for preparing this massive migration to our new clair+c2py binding framework. I have pushed a number of smaller updates, fixes and clean-ups. Please have a look once. From my side this is now ready for merge.

PR #1006, opened 2026-04-28
[op] U_matrix:spherical_to_cubic add l=3 support

the-hampel:
This commit adds l=3 support for spherical_to_cubic for convention=triqs,vasp,qe, and wannier90 . Previously only triqs was supported.

I also added the trivial l=0 code path.

For Wien2k this seems to be not so trivial since the case.cf_f_mm2 only specific to C2v symmetry and no generic transformation is possible. Users have to do this manually and call via convention='triqs' then.

This also implements a trivial unitary check for the transformation.

This fixes an issue raised here: https://github.com/TRIQS/solid_dmft/issues/114

### DIFF
--- a/python/triqs/operators/util/U_matrix.py
+++ b/python/triqs/operators/util/U_matrix.py
@@ -405,33 +405,28 @@ def spherical_to_cubic(l, convention='triqs'):
             T[3,1] = 1.0/sqrt(2);   T[3,3] = -1.0/sqrt(2)
             T[4,0] = 1.0/sqrt(2);   T[4,4] = 1.0/sqrt(2)
     elif l == 3:
+        # l=3 cubic orderings used here:
+        # triqs:     ("x(x^2-3y^2)", "z(x^2-y^2)", "xz^2", "z^3", "yz^2", "xyz", "y(3x^2-y^2)")
+        # vasp:      ("y(3x^2-y^2)", "xyz", "yz^2", "z^3", "xz^2", "z(x^2-y^2)", "x(x^2-3y^2)")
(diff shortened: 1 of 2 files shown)


jcoulter12:
Hi Alex, 

The PR seems mostly good -- two minor suggestions, one from me and one from Nils. Let me know what you think.

1) Would you mind adding a small comment to this test about what it is testing? Is it a totally trivial case, or is it sufficient to cover the general use of spherical_to_cubic (and if so, can we think of another small test?) 
2) Maybe instead of spelling out the T matrix three times, it's possible to define it once and then permute it depending on the DFT code? 

Thanks, 
Jenny

the-hampel:
Hi Jenny, 

thank you for taking a look at this. Yes I also thought about using some nicer way to do the permutations between the codes. I a mostly refrained from this to stay with the current scheme (l=1 and l=2 code). Anyway I agree that this is nicer and less error prone. 

Regarding tests: The current test is really a baseline check but not really sophisticated. It just check if the permutations applied will still give a unitary matrix. I added a comment. I also check now if the permutations work correctly. 

But honestly these are just simple tests to check whether the code still runs. Similar for what is done for l=2 and l=1. If you have any other idea let me know. It is not great but the best that came to my mind. 

Best,
Alex

jcoulter12:
Hi Alex, 

Looks good, thank you for making these extra changes. 

On additional tests, I guess the best way to test is really by having a couple of material examples with f orbitals from DFT codes, which would need to be tested in one of the material-specific apps rather than the main library, anyway. Hopefully there's some use in the solid_dmft tests. 

Either way for now, it's good there's something describing the test for now. 
The issue that keeps the CI from passing is something Nils is looking into, I think -- so we'll wait for that fix before proceeding. Lots going on with the library this week. :) 

Thanks, 
Jenny

jcoulter12 (approved):

PR #1008, opened 2026-05-11
Fix recompilation python modules

parcollet:
Fix multiple recompilations of c2py modules.


Wentzell:
Thank you for pointing this out. This is now fixed by the merge of #1010

PR #1009, opened 2026-05-12
[test] add tests for SemiCircular testing mu!=0 and fix after #1003

### LINKED ISSUE #1007: SemiCircular(D, chem_potential) produces wrong densities — regression introduced by PR #1003
`SemiCircular(D, mu)` with a non-zero chemical potential produces completely wrong Green's function values and densities since PR #1003 was merged. This led to test fails in solid_dmft. I am not sure if this is an intended site effect. I understand that #1003 was meant to make Im axis eval faster but somehow the real part of the argument is now lost. The function `SemiCircular` has two arguments: half-bandwidth, and mu. So this should work or the wrapper removed I think.

Minimal reproducer:
```
from triqs.gf import Gf, MeshImFreq
from triqs.gf.descriptors import SemiCircular

mesh = MeshImFreq(beta=40., statistic='Fermion', n_iw=1025)

# mu=0: symmetric DOS, expect occupation = 0.5
g = Gf(mesh=mesh, target_shape=[1, 1])
g << SemiCircular(2.0, 0.0)
print(f"mu=0:  n = {g.density().real[0,0]:.6f}  (expect 0.500000)")

# mu=1: DOS centre shifted below Fermi, expect occupation ~ 0.8044
g << SemiCircular(2.0, 1.0)
print(f"mu=1:  n = {g.density().real[0,0]:.6f}  (expect ~0.804398)")

# mu=-1: expect occupation ~ 0.1956
g << SemiCircular(2.0, -1.0)
print(f"mu=-1: n = {g.density().real[0,0]:.6f}  (expect ~0.195602)")
```

Output before PR #1003 (correct):
```
mu=0:  n = 0.500000  (expect 0.500000)
mu=1:  n = 0.804398  (expect ~0.804398)
mu=-1: n = 0.195602  (expect ~0.195602)
```

Output after PR #1003 (broken):
```
mu=0:  n = 0.500000  (expect 0.500000)
mu=1:  n = 107.9...  (expect ~0.804398)
mu=-1: n = -106.9... (expect ~0.195602)
```

I think this is because PR #1003 introduced semicirc.py and refactored descriptors.py::SemiCircular to call g_semicirc_iw(iw, D). The descriptor passes z = iω_n + μ , but g_semicirc_iw was written assuming a purely imaginary argument:

```
# semicirc.py — broken for complex z
def g_semicirc_iw(iw, D):
    w = iw.imag          # discards the real part μ from the sqrt/sign
    return (2.0 / D**2) * (iw - 1j * np.sign(w) * np.sqrt(D**2 + w**2))
    #                       ^^^ keeps μ here, but not in the sqrt
    #       net effect: adds a constant offset of +2μ/D² to every G(iω_n)
```

Before PR #1003, descriptors.py used cmath.sqrt() inline, which correctly handled the complex argument. The refactor lost that capability and worse does not check if the argument is purely imaginary. So we have to either restore old capability or check if all arguments are purely imaginary.

One way to fix this is to check if the argument is purely imaginary and then take a short cut, and if not go the "old" route: 
```
def g_semicirc_iw(iw, D):
    iw = np.asarray(iw, dtype=complex)
    if np.all(iw.real == 0):
        # Fast path for purely imaginary Matsubara frequencies
        w = iw.imag
        return (2.0 / D**2) * (iw - 1j * np.sign(w) * np.sqrt(D**2 + w**2))
    # General formula for complex arguments (e.g. shifted by chem_potential)
    sqrt_val = np.sqrt(iw**2 - D**2 + 0j)
    nonzero = iw.imag != 0
    flip = nonzero & (np.sign(sqrt_val.imag) != np.sign(iw.imag))
    sqrt_val = np.where(flip, -sqrt_val, sqrt_val)
    return (2.0 / D**2) * (iw - sqrt_val)
```

I have a commit ready with this one-function fix plus a regression test added to test/python/base/gf_semicirc_tau.py that covers SemiCircular(D, μ) for μ > 0, μ = 0, μ < 0 ... Let me know if I should open a PR or you would like to fix this on your own.

Best,
Alex

the-hampel:
closes #1007

this PR adds tests for the SemiCircular wrapper in triqs to test also non-zero chemical potential. There was a regression undetected in PR #1003 that has to be fixed. 

This is split into two commits. 1) add proper testing 2) add a preliminary fix. 

Note: even though I carefully checked and tested this code and reviewed myself  I wrote this with Claude. So please recheck. I tried to make the changes as minimal as possible. Maybe there is a better alternative route to this. Feel free to change.

### DIFF
--- a/python/triqs/gfs/descriptors.py
+++ b/python/triqs/gfs/descriptors.py
@@ -23,7 +23,7 @@ r""" """
 
 from .descriptor_base import *
 from triqs.mesh import MeshImFreq, MeshDLRImFreq, MeshReFreq, MeshReFreqPts, MeshReFreqLog, MeshImTime, MeshDLRImTime
-from .semicirc import g_semicirc_iw, g_semicirc_w, g_semicirc_tau
+from .semicirc import g_semicirc_z, g_semicirc_w, g_semicirc_tau
 import warnings
 
 #######################################
@@ -89,7 +89,7 @@ semicircle
         mu = self.chem_potential
         Id = 1. if len(G.target_shape) == 0 else numpy.identity(G.target_shape[0])
         if type(G.mesh) in [MeshImFreq, MeshDLRImFreq]:
-            vals = g_semicirc_iw(G.mesh.values() + mu, D)
+            vals = g_semicirc_z(G.mesh.values() + mu, D)
         elif type(G.mesh) in [MeshReFreq, MeshReFreqPts, MeshReFreqLog]:
             vals = g_semicirc_w(G.mesh.values().real + mu, D)
         elif type(G.mesh) in [MeshImTime, MeshDLRImTime]:
--- a/python/triqs/gfs/semicirc.py
+++ b/python/triqs/gfs/semicirc.py
@@ -26,14 +26,27 @@ from scipy.integrate import quad
 def g_semicirc_iw(iw, D):
     r"""Semi-circular Green's function on the Matsubara axis.
 
+    For purely imaginary arguments :math:`i\omega_n`:
+
     .. math::
         G(i\omega) = \frac{2}{D^2}\bigl(i\omega
             - i\,\mathrm{sign}(\omega)\,\sqrt{D^2 + \omega^2}\bigr)
 
+    For general complex arguments :math:`z = i\omega_n + \mu` (used when a
+    chemical potential shift is applied via :class:`SemiCircular`):
+
+    .. math::
+        G(z) = \frac{2}{D^2}\bigl(z - \sqrt{z^2 - D^2}\bigr)
+
+    where the branch of the square root is chosen so that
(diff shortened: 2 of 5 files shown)


jasonkaye:
Thanks for this @the-hampel. I replaced g_semicirc_iw with a general g_semicirc_z which uses a simple formula and should work anywhere in the complex plane except [-D, D]---in particular, at z = iw + mu. Please double check that this is satisfactory, and that tests still pass. Does this implementation lose anything compared to yours?

@Wentzell Please check that the tests added by this PR are satisfactory.

the-hampel:
@jasonkaye thank you! Looks very nice now. :-) Much more advanced than my brute force way. Locally all tests pass. I just rebased on current unstable to see if that fixes the CI now. Otherwise this is good to merge from my side.

PR #1010, opened 2026-05-13
Fix multiple recompilation of python modules (again)

Thoemi09:
This is basically the same as https://github.com/TRIQS/triqs/pull/1008, just a bit simplified and rebased onto unstable.


Wentzell (approved):
Thank you @Thoemi09 for this fix

PR #1013, opened 2026-06-02
[jenkins] migration to new k8s-system

dylex:
This is a fairly major overhaul to get triqs to build on the new system. This build is now passing [here](https://jenkins-new.flatironinstitute.org/job/CCQ/job/TRIQS/job/triqs/job/jenkins/).

The biggest change is a move away from doing builds inside Dockerfiles, but rather using the Dockerfile only for the base environment, and then installing triqs (and other applications) in a shared NFS mount. This make linux builds work more like the Mac builds were working, and allows more flexibility to build and run tests across multiple nodes.

I've also tried to generalize the jenkins setup itself using a github [organization folder](https://jenkins-new.flatironinstitute.org/job/CCQ/job/TRIQS/) that will automatically discover any repos in the TRIQS organization that have Jenkinsfiles. We can tweak the template configuration if needed.

In order to avoid doing broken builds on the new system, I also moved the Jenkinsfile to a new path.

The documentation update is working, but I haven't tested or done anything with the packaging repo yet.


dylex:
This is now setup to run in parallel on the new jenkins system, using separate config files in the `jenkins` directory, while not touching anything else. Pushing docs and publishing is explicitly disabled.

jenkins/Jenkinsfile
```diff
@@ -0,0 +1,205 @@
...
+catchError {
+  parallel platforms
+
+  if (keepInstall) {
+    /* Publish results */
+    stage("publish") {
+      timeout(time: 10, unit: 'MINUTES') {
+        runPod(tag: documentationPlatform,
+          mounts: mounts) {
+          checkout scm
```
dylex: This `checkout scm` is not necessary and will cause a warning -- `runPod` already does it in the correct context.

PR #1014, opened 2026-06-03
Fix UB issue in lin- and log_binning

Thoemi09:
- avoid divide-by-zero which triggers UBSAN


Wentzell (approved):
Looking good, I applied some minor clean-up.

PR #1015, opened 2026-06-10
DLR mesh stability

Wentzell:
Check dlr mesh stability for different build environments by comparing against reference data.


Wentzell:
Merged in 9bed4fae

PR #1016, opened 2026-06-10
Add triqs.utility.utilities python module for C++ code in triqs/utility

Thoemi09:
- porting cthyb to clair + c2py requires `time_pt` to be wrapped


Wentzell (approved):
Looks good, thank you @Thoemi09

PR #1020, opened 2026-06-26
[jenkins] make docker builds more efficient by avoiding extra checkouts

dylex:
This is just an optimization, but probably saves a couple minutes. I can make the same change on app4triqs.


Wentzell (approved):
Thank you @dylex for this improvement.

PR #1023, opened 2026-07-30
Bugfix for bz_integration for edge case where mu and H(kx,ky,kz = 0) 

jcoulter12:
In an edge case where both mu and H(k) at specifically the kx,ky,kz = 0 point, gloc or find_chemical_potential fail as  nda::linalg::inv_in_place error. This originated from the deduce_dim_from_expression function, which was filling in placeholders with kx,ky,kz = 0, w=0 to evaluate the lazy expression to be integrated and find it's dimension. 
This has been fixed by using the first mesh point on which a function will be evaluated instead of w=0.

Additional minor changes: 
-- An unused integrate_ptr function was removed. It was never instantiated and had implementation issues anyway -- it was designed it seems, to output integrated k quantities to a gf container. We can add this later if it seems it's needed. 
-- There was a problem in verbose mode with printing the remaining number of unconverged frequency points. 
-- Upon suggestion from @dabramovitch, we changed the meaning of k_grid_max. Previously, we ran *up to* but not including the max grid. This would change the behavior to be more intuitive, running also the maximum specified grid density. As in, if k_grid_max = [10,10,10], and k_grid incremented in units of 1, before the last k_grid run would be 9,9,9. Now it will also run the 10,10,10 one before stopping. 

Thanks to @dabramovitch for reporting and helping to locate this issue.

### DIFF
--- a/c++/triqs/experimental/lattice/bz_integrators.hpp
+++ b/c++/triqs/experimental/lattice/bz_integrators.hpp
@@ -79,10 +79,21 @@ namespace triqs::experimental::lattice {
     // helper to determine the return container dimension
     template <typename T>
     int deduce_dim_from_expression(auto const &f_kw, T const &w) {
+
       namespace ph  = placeholders;
-      // need to use w here and not 0, as this can result in the expression becoming 1/0
-      auto f_w_temp = eval(f_kw, ph::kx = 0., ph::ky = 0., ph::kz = 0., ph::w = w);
+
+      // check that there will not be an nda::linalg::inv error due to a div by zero at this k, w
+      auto f_w_temp = [&](){
+        try {
+            // need to use w here and not 0, as this can result in the expression becoming 1/0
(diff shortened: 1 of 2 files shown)


jcoulter12:
I realized there may be one more even narrower edge case to handle, maybe delay review until I've looked at it.

jcoulter12:
This should be finished and could be merged -- I have added an additional check in this deduce_dim function to throw a more helpful error in case someone makes a real frequency mesh GF which includes a 1/0. 

However, this case (real frequency mesh including w=0, mu=0, Ek=0, sigma=0) will still cause issues in other parts of these functions if someone does this. We can't add these try catch blocks with a more helpful error message everywhere without compromising on performance (the deduce_dim function is called only once per integration, so not an issue there). Because this is a very specific and unphysical case, I'm currently advocating to just allow nda::linalg::inv_in_place to throw a runtime error. A slightly educated user should be able to figure this out. 

Thanks for the patience on this.

